PR #9, opened 2020-04-03
updated .drone.yml

kdeng3849:


.drone.yml
```diff
@@ -30,28 +62,20 @@ pipeline:
...
       - CGO_ENABLED=0 go build
 
-  publish:
-    group: publish
-    image: plugins/docker
-    registry: quay.io
+  publish_pr_hegel:
+    <<: *publish_pr_tmpl
+    context: .
+    dockerfile: Dockerfile
```
mmlb, line 68: you can drop `context` and `dockerfile` from all the locations, this is the default already. I added to tink just so it all looks the same.

PR #12, opened 2020-05-21
add prometheus metrics

mikemrm:


grpc_server.go
```diff
@@ -86,11 +75,22 @@ func (s *server) Get(ctx context.Context, in *hegel.GetRequest) (*hegel.GetRespo
 }
 
 func (s *server) Subscribe(in *hegel.SubscribeRequest, stream hegel.Hegel_SubscribeServer) error {
-	active_subscription.Inc()
-	defer active_subscription.Dec()
+	metrics.TotalSubscriptions.Inc()
+	metrics.Subscriptions.WithLabelValues("initializing").Inc()
+	initError := func(err error) error {
+		s.log.Error(err)
```
mmlb, line 81: would be nice to have the ip address logged too once its been extracted from the request

mikemrm: Good point

grpc_server.go
```diff
@@ -86,27 +75,43 @@ func (s *server) Get(ctx context.Context, in *hegel.GetRequest) (*hegel.GetRespo
...
+	logger := s.log.With("op", "subscribe")
+
+	initError := func(err error) error {
+		logger.Error(err)
+		metrics.Subscriptions.WithLabelValues("initializing").Dec()
+		labels := prometheus.Labels{
+			"op":    "subscribe",
+			"state": "initializing",
+		}
+		metrics.Errors.With(labels).Inc()
```
mmlb, line 90: why the `labels` var? Don't really need to change if you don't want though.

mikemrm: ah, you're right. Would be nice to keep it consistent.

PR #13, opened 2020-06-02
Add compatibility with tink

kdeng3849:
(Sorry I made a branch directly off of the main repo...)

Changes:
- Added an envvar `HARDWARE_DATA_MODEL` to switch between the old cacher model and the new tinkerbell model
  - if envvar is unset/empty/anything other than `tinkerbell`, it will default to using the old cacher data model
- Abstracted `exportedHardware` among others to take into account new data model

TODO:
- [ ] remove the `replace` in `go.mod` pointing to my fork (after the pr in tink gets merged)
- [ ] test `Subscribe` method (honestly, not sure how)
- [ ] health check tink when creating new tink client?  https://github.com/tinkerbell/hegel/blob/ec7c03373f17ec839c5bc670803637dfe3f2ebd3/main.go#L145

Related PRs:
boots: https://github.com/tinkerbell/boots/pull/27
tink: https://github.com/tinkerbell/tink/pull/64


data_fetch.go
```diff
@@ -0,0 +1,183 @@
...
+	"context"
+	"os"
+)
+
+var fetcher dataFetch
+
+const (
+	cacherDataModel = `
+	{
+	  "id": "8978e7d4-1a55-4845-8a66-a5259236b104",
```
yetang-equinix, line 13: why this id is hard coded here?

kdeng3849: This is a mock of what would be returned from cacher so that in the tests we don't have to actually make calls to cacher. I'm working on moving these constants into the actual test file so it's not confusing like so.

yetang-equinix: if just for testing, fine for now.

PR #20, opened 2020-07-24
ENG-7511 custom endpoints

kdeng3849:
## Description

Allows users to freely configure their own endpoints and what parts of the hardware data the endpoints are to return.

## Why is this needed

Since we allow users to freely input metadata, there is no way of knowing the structure of their metadata and what parts they want to expose. Currently, we just have a single endpoint `/metadata` that returns all of metadata inside the hardware data. But with this, users will have more control.

Fixes: #

## How Has This Been Tested?

- unit tests
- manual test using tinkerbell setup

No migration steps are needed, but there is now an added optional `CUSTOM_ENDPOINTS` environment variable that is a json-formatted string of a map with keys being the endpoint patterns and values being the filter used to parse the hardware data. The filters are `jq` filters.
If the `CUSTOM_ENDPOINTS` envvar is _unset_, the default will be `{"/metadata":".metadata"}`, which sets up a single endpoint `/metadata` and returns all of metadata (what hegel currently does).

Example:
Let's say your `CUSTOM_ENDPOINTS` has the value `{"/facility":".metadata.facility.facility_code", "/all":".", "/userdata":".metadata.userdata"}` and the following hardware is stored in the tink database:

```
{
  "id": "0eba0bf8-3772-4b4a-ab9f-6ebe93b90a94",
  "metadata": {
    "userdata": "#!/bin/bash\n\necho \"Hello world!\"",
    "facility": {
      "facility_code": "ewr1",
      "plan_slug": "c2.medium.x86",
      "plan_version_slug": ""
    },
    "state": ""
  },
  "network": {
    ...
  }
}

```

This means, you will have 3 endpoints: `/facility`, `/all`, and `/userdata`.
- `/facility` returns `ewr1`
- `/all` returns 
```
{
   "userdata":"#!/bin/bash\n\necho \"Hello world!\"",
   "facility":{
      "facility_code":"ewr1",
      "plan_slug":"c2.medium.x86",
      "plan_version_slug":""
   },
   "state":""
}
```
- `/userdata` returns 
```
#!/bin/bash

echo "Hello world!"
```
## Checklist:

I have:

- [x] added unit or e2e tests


http_handlers.go
```diff
@@ -52,25 +54,45 @@ func healthCheckHandler(w http.ResponseWriter, r *http.Request) {
...
-				return
-			}
-			w.WriteHeader(http.StatusOK)
-			w.Header().Set("Content-Type", "application/json")
-			_, err = w.Write(ehw)
-			if err != nil {
-				logger.Error(err, "failed to write Metadata")
+func getMetadata(filter string) http.HandlerFunc {
+	return func(w http.ResponseWriter, r *http.Request) {
+		if r.Method == "GET" {
```
nathangoulding, line 59: Can we flip this around to return if r.Method != "GET" so we're indenting less?

http_handlers.go
```diff
@@ -52,25 +54,45 @@ func healthCheckHandler(w http.ResponseWriter, r *http.Request) {
...
+			logger.Debug("Calling getMetadata ")
+			userIP := getIPFromRequest(r)
+			if userIP != "" {
```
nathangoulding, line 62: Same here, return if it's empty and bring the indenting in a bit

main.go
```diff
@@ -250,7 +252,10 @@ func main() {
 	http.Handle("/metrics", promhttp.Handler())
 	http.HandleFunc("/_packet/healthcheck", healthCheckHandler)
 	http.HandleFunc("/_packet/version", versionHandler)
-	http.HandleFunc("/metadata", getMetadata)
+	err = registerCustomEndpoints()
+	if err != nil {
+		logger.Error(err, "could not register custom endpoints")
```
nathangoulding, line 257: This would be a good place to `fatal` I think.

PR #22, opened 2020-07-30
ENG-7133 add EC2 endpoint

kdeng3849:
## Description

Adds the EC2 endpoint `/2009-04-04`.

## Why is this needed

There has been an effort to have Hegel succeed Kant, and one of the endpoints that Kant has (and Hegel doesn't) is the `/2009-04-04` endpoint. 

Fixes: #

## How Has This Been Tested?

- Integration tests
- Manual test with tinkerbell setup

No migration steps needed.

## Checklist:

I have:

- [x] added unit or e2e tests

### DIFF
--- a/grpc_server.go
+++ b/grpc_server.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/json"
 	"errors"
+	"fmt"
 	"io"
 	"math"
 	"net"
@@ -177,7 +178,7 @@ func exportHardware(hw []byte) ([]byte, error) {
 }
 
 func filterMetadata(hw []byte, filter string) ([]byte, error) {
-	var result interface{}
+	var result string
 	query, err := gojq.Parse(filter)
 	if err != nil {
 		return nil, err
@@ -196,16 +197,19 @@ func filterMetadata(hw []byte, filter string) ([]byte, error) {
 		if err, ok := v.(error); ok {
 			return nil, err
 		}
-		result = v
+
+		if vString, ok := v.(string); ok { // if already a string, don't marshal
+			result = fmt.Sprintln(result + vString) // append v to result
+		} else if v != nil { // if nil, don't marshal (json.Marshal(nil) returns "null")
+			vByte, err := json.Marshal(v)
+			if err != nil {
+				return nil, err
+			}
+			result = fmt.Sprintln(result + string(vByte)) // append v to result
+		}
 	}
 
-	if resultString, ok := result.(string); ok { // if already a string, don't marshal
-		return []byte(resultString), nil
-	}
-	if result != nil { // if nil, don't marshal (json.Marshal(nil) returns "null")
-		return json.Marshal(result)
-	}
-	return nil, nil
+	return []byte(strings.TrimSuffix(result, "\n")), nil
 }
 
 // UnmarshalJSON implements the json.Unmarshaler interface for custom unmarshalling of exportedHardwareCacher
--- a/grpc_server.go
+++ b/grpc_server.go
@@ -199,13 +199,13 @@ func filterMetadata(hw []byte, filter string) ([]byte, error) {
 		}
 
 		if vString, ok := v.(string); ok { // if already a string, don't marshal
-			result = fmt.Sprintln(result + vString) // append v to result
+			result += fmt.Sprintln(vString) // append v to result
 		} else if v != nil { // if nil, don't marshal (json.Marshal(nil) returns "null")
 			vByte, err := json.Marshal(v)
 			if err != nil {
 				return nil, err
 			}
-			result = fmt.Sprintln(result + string(vByte)) // append v to result
+			result += fmt.Sprintln(string(vByte)) // append v to result
 		}
 	}
 
--- a/grpc_server.go
+++ b/grpc_server.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"bytes"
 	"context"
 	"encoding/json"
 	"errors"
@@ -177,7 +178,7 @@ func exportHardware(hw []byte) ([]byte, error) {
 }
 
 func filterMetadata(hw []byte, filter string) ([]byte, error) {
-	var result string
+	var result bytes.Buffer
 	query, err := gojq.Parse(filter)
 	if err != nil {
 		return nil, err
@@ -198,17 +199,18 @@ func filterMetadata(hw []byte, filter string) ([]byte, error) {
 		}
 
 		if vString, ok := v.(string); ok { // if already a string, don't marshal
-			result += vString + "\n" // append v to result
+			result.WriteString(vString + "\n") // append v to result
 		} else if v != nil { // if nil, don't marshal (json.Marshal(nil) returns "null")
 			vByte, err := json.Marshal(v)
 			if err != nil {
 				return nil, err
 			}
-			result += string(vByte) + "\n" // append v to result
+			result.Write(vByte) // append v to result
+			result.WriteRune('\n')
 		}
 	}
 
-	return []byte(strings.TrimSuffix(result, "\n")), nil
+	return bytes.TrimSuffix(result.Bytes(), []byte("\n")), nil
 }
 
 // UnmarshalJSON implements the json.Unmarshaler interface for custom unmarshalling of exportedHardwareCacher
--- a/grpc_server.go
+++ b/grpc_server.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"encoding/json"
 	"errors"
-	"fmt"
 	"io"
 	"math"
 	"net"
@@ -199,13 +198,13 @@ func filterMetadata(hw []byte, filter string) ([]byte, error) {
 		}
 
 		if vString, ok := v.(string); ok { // if already a string, don't marshal
-			result += fmt.Sprintln(vString) // append v to result
+			result += vString + "\n" // append v to result
 		} else if v != nil { // if nil, don't marshal (json.Marshal(nil) returns "null")
 			vByte, err := json.Marshal(v)
 			if err != nil {
 				return nil, err
 			}
-			result += fmt.Sprintln(string(vByte)) // append v to result
+			result += string(vByte) + "\n" // append v to result
 		}
 	}
 
--- a/grpc_server.go
+++ b/grpc_server.go
@@ -415,7 +415,7 @@ func getByIP(ctx context.Context, s *server, userIP string) ([]byte, error) {
 		req := &tink.GetRequest{
 			Ip: userIP,
 		}
-		resp, err := s.hardwareClient.ByIP(ctx, req) // use wrapper?
+		resp, err := s.hardwareClient.ByIP(ctx, req)
 
 		if err != nil {
 			return nil, err
--- a/grpc_server_test.go
+++ b/grpc_server_test.go
@@ -20,7 +20,7 @@ func TestGetByIPCacher(t *testing.T) {
 			log:            logger,
 			hardwareClient: hardwareGetterMock{test.json},
 		}
-		ehw, err := getByIP(context.Background(), hegelTestServer, test.remote) // returns hardware data as []byte
+		ehw, err := getByIP(context.Background(), hegelTestServer, mockUserIP) // returns hardware data as []byte
 		if err != nil {
 			t.Fatal("unexpected error while getting hardware by ip:", err)
 		}
@@ -82,7 +82,7 @@ func TestGetByIPTinkerbell(t *testing.T) {
 			log:            logger,
 			hardwareClient: hardwareGetterMock{test.json},
 		}
-		ehw, err := getByIP(context.Background(), hegelTestServer, test.remote) // returns hardware data as []byte
+		ehw, err := getByIP(context.Background(), hegelTestServer, mockUserIP) // returns hardware data as []byte
 		if err != nil {
 			if err.Error() != test.error {
 				t.Fatalf("unexpected error in getByIP, want: %v, got: %v\n", test.error, err.Error())
@@ -144,7 +144,6 @@ func TestGetByIPTinkerbell(t *testing.T) {
 
 var cacherGrpcTests = map[string]struct {
 	id               string
-	remote           string
 	state            string
 	facility         string
 	mac              string
@@ -159,7 +158,6 @@ var cacherGrpcTests = map[string]struct {
 	error            string
 }{
 	"cacher": {
-		remote:           "192.168.1.5",
 		state:            "provisioning",
 		facility:         "onprem",
 		mac:              "98:03:9b:48:de:bc",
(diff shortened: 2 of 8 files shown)


grpc_server.go
```diff
@@ -196,16 +197,19 @@ func filterMetadata(hw []byte, filter string) ([]byte, error) {
...
-		result = v
-	}
 
-	if resultString, ok := result.(string); ok { // if already a string, don't marshal
-		return []byte(resultString), nil
-	}
-	if result != nil { // if nil, don't marshal (json.Marshal(nil) returns "null")
-		return json.Marshal(result)
+		if vString, ok := v.(string); ok { // if already a string, don't marshal
+			result = fmt.Sprintln(result + vString) // append v to result
```
mmlb, line 202: this doesn't really make sense, since first you are concatenating both strings and then basically just returning it but with an extra new-line (from the `ln` family) which then gets stripped off when returning at the end.

These calls can all just be `result += vString` or `result += string(vByte)` followed with just `return []byte(result)`

http_handlers.go
```diff
@@ -102,6 +140,106 @@ func getMetadata(filter string) http.HandlerFunc {
 	}
 }
 
+func ec2Handler(w http.ResponseWriter, r *http.Request) {
+	if r.Method != "GET" {
+		return
```
mmlb, line 145: why don't we return a `StatusMethodNotAllowed` error to the client?

http_handlers.go
```diff
@@ -102,6 +140,106 @@ func getMetadata(filter string) http.HandlerFunc {
...
+	}
+
+	logger.Debug("Calling ec2Handler ")
+	userIP := getIPFromRequest(r)
+	if userIP != "" {
```
mmlb, line 150: return early if `userIP == ""` and avoid a level of indentation

kdeng3849: Oops, forgot to change this. Will do now.

grpc_server.go
kdeng3849: The new line is really for the list values returned as separate results (i.e. tags and ssh keys). If I don't add a new line at the end of each iteration, `["tag1", "tag2"]` will become `tag1tag2` as the result, instead of 
```
tag1
tag2
```
I remove the new line at the very end (outside of the loop) so it matches the behavior in Kant, which also doesn't return a new line at the end of the response.

For _some reason_, I thought `+=` was not a thing in golang.... I'll reformat that..

http_handlers.go
```diff
@@ -102,6 +141,109 @@ func getMetadata(filter string) http.HandlerFunc {
...
+			logger.Error(err, "Error in filtering metadata: ")
+		}
+	} else if submenu, ok := res.(map[string]interface{}); ok {
+		var keys []string
+		for item := range submenu {
+			switch item {
+			case "_base": // _base is only used to keep track of the base filter, not a metadata item
+				continue
+			case "spot": // list only if instance is spot
+				resp, err := filterMetadata(hw, ".metadata.instance.spot")
```
kdeng3849, line 189: Is this okay to be hardcoded? Should it refer to the `ec2Filters` map instead? 

**Pros:** changes to the `spot` filter inside `ec2Filters` will automatically change the filter here as well  
**Cons:** To build that same path, I would need to do `submenu["_base"] + submenu[item]["_base"]`, which is harder to read.

http_handlers_test.go
```diff
@@ -258,6 +299,13 @@ var registerEndpointTests = map[string]struct {
 		status:          200,
 		json:            tinkerbellDataModel,
 	},
+	"single custom endpoint, non-metadata": {
+		customEndpoints: `{"/id": ".id"}`,
+		url:             "/id",
+		remote:          "192.168.1.5",
```
DavidHwu, line 305: hard coded test IP... does it make sense to have a const here for self documenting code.  What does this remote IP reflect?

http_handlers.go
```diff
@@ -102,6 +141,109 @@ func getMetadata(filter string) http.HandlerFunc {
...
+	w.Header().Set("Content-Type", "application/json")
+	_, err = w.Write(resp)
+	if err != nil {
+		logger.Error(err, "failed to write Metadata")
+	}
+}
+
+// processEC2Query returns either a specific filter (used to parse hardware data for the value of a specific field),
+// or a map of filters (used for printing out its keys)
+func processEC2Query(query string) (interface{}, error) {
```
DavidHwu, line 218: does it make sense to do a unit test around this function?

http_handlers_test.go
```diff
@@ -171,6 +174,40 @@ func TestRegisterEndpoints(t *testing.T) {
 	}
 }
 
+func TestEC2Endpoint(t *testing.T) {
```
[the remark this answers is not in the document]
kdeng3849: yeah, drone runs all the tests on push

http_handlers_test.go
kdeng3849: because hegel looks up the ip from the request's `RemoteAddr` (to then query cacher/tink for the hardware data), this field is to manually set the `request.RemoteAddr` in the tests to make pretend that's where the request came from.

code for reference:
- hegel looking up the ip from the request
https://github.com/tinkerbell/hegel/blob/415e1308915b3fec2f7da7543918b895a9c6452a/http_handlers.go#L247
- setting the remote addr in the tests
https://github.com/tinkerbell/hegel/blob/415e1308915b3fec2f7da7543918b895a9c6452a/http_handlers_test.go#L73

http_handlers.go
kdeng3849: Sure, I can add one in

http_handlers_test.go
DavidHwu: https://itnext.io/tips-for-writing-self-documenting-code-e54a15e9de2

GO code can seem esoteric as the origins comes from C/C++, thus the code can be terse (even with the naming conventions).  Unlike other languages (ie. PEP20 focuses on explicit over implicit) and have specifics to address these in the language and linter themselves.

With any good code, bugs, misunderstandings and mistakes can be made with good software practice starting with good self documenting code.

https://itnext.io/tips-for-writing-self-documenting-code-e54a15e9de2

Reflects on key principles for self documenting code:

Being readable vs. clever able.
This reduces mistakes, misunderstandings and bugs introduced as such
"Being a good developer doesn’t mean writing the cleverest code, it means being a good teammate"
Use clear variable names
Use small named functions
No magic numbers... extrapolated also to no magic commands (ie. complex SQL commands).
Remember think of your team and the story you are conveying wrt your intent.

These are all excerpts from the PRI side for Go coding standards we have in place
https://wiki.corp.equinix.com/confluence/display/ES/GO+Coding+Standards

kdeng3849: Ohh, I see. I didn't know that was the name for this. 

I don't think it makes sense to have a constant for a test case value because these values are chosen arbitrarily. The only significance `192.168.1.5` has is that it is the IP used in that specific mock hardware json. To create a const for it might actually mislead others into thinking it's more significant than it is.  
With the way these tests are set up, the name `remote`  _is_ meant to describe what `192.168.1.5` reflects, but I must admit, `remote` was a poorly chosen name on my part. So what I think I should do is rename `remote` to `userIP` so it's clear that `192.168.1.5` (or any other value assigned to that field) refers to the mock user's IP.

Is that reasonable?

kdeng3849: Actually, it turns out that the `remote` ip could literally be anything and the mocks would still be returned (the ip isn't actually being matched/checked). So I ended up just creating a constant.

http_handlers.go
```diff
@@ -102,6 +141,113 @@ func getMetadata(filter string) http.HandlerFunc {
...
+			logger.Error(err, "Error in filtering metadata: ")
+		}
+	} else if submenu, ok := res.(map[string]interface{}); ok {
+		var keys []string
+		for item := range submenu {
+			switch item {
+			case "_base": // _base is only used to keep track of the base filter, not a metadata item
+				continue
+			case "spot": // list only if instance is spot
+				spotFilter := fmt.Sprint(submenu["_base"], submenu[item].(map[string]interface{})["_base"])
```
mmlb, line 189: we don't need any of these `fmt.Sprint` calls over `arg1 + arg2` since they are both strings already.

grpc_server.go
kdeng3849: Oh, instead of using fmt.Sprint? Is there a reason why?

http_handlers.go
```diff
@@ -102,6 +141,109 @@ func getMetadata(filter string) http.HandlerFunc {
...
+		logger.Info("Error in finding hardware: ", err)
+		w.WriteHeader(http.StatusInternalServerError)
+		return
+	}
+
+	res, err := processEC2Query(r.URL.Path)
+	if err != nil {
+		logger.Error(err)
+		w.WriteHeader(http.StatusNotFound)
+		_, err := w.Write([]byte("404 not found"))
```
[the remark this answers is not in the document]
kdeng3849: This is actually a different error code + message compared to Kant, since the original was really vague: `{"status":500,"message":"Something went wrong..."}`. 
Do you have a suggestion for this?

http_handlers.go
```diff
@@ -102,6 +141,109 @@ func getMetadata(filter string) http.HandlerFunc {
...
+				if err != nil {
+					logger.Info("Error in filtering metadata: ", err)
+				}
+				if string(resp) != "" {
+					keys = append(keys, item)
+				}
+			default:
+				keys = append(keys, item)
+			}
+		}
```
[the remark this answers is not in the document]
kdeng3849: What do you mean by branch? can you give an example?

http_handlers.go
kdeng3849: I'm not sure if this makes a difference, but `submenu["_base"]` is actually an `interface{}` so if I use the `+` to concat, I would need to do type assertions:
```
spotFilter := submenu["_base"].(string) + submenu[item].(map[string]interface{})["_base"].(string)
```
which looked really cluttered to me. 
I guess, how does `sprint` compare to type assertions in terms of efficiency?

grpc_server.go
```diff
@@ -177,7 +178,7 @@ func exportHardware(hw []byte) ([]byte, error) {
 }
 
 func filterMetadata(hw []byte, filter string) ([]byte, error) {
-	var result interface{}
+	var result string
```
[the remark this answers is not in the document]
kdeng3849: okay yeah, I'll try that

grpc_server_test.go
```diff
@@ -20,7 +20,7 @@ func TestGetByIPCacher(t *testing.T) {
 			log:            logger,
 			hardwareClient: hardwareGetterMock{test.json},
 		}
-		ehw, err := getByIP(context.Background(), hegelTestServer, test.remote) // returns hardware data as []byte
+		ehw, err := getByIP(context.Background(), hegelTestServer, mockUserIP) // returns hardware data as []byte
```
DavidHwu, line 23: nice thanks @kdeng3849  I understand 👍

http_handlers.go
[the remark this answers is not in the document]
DavidHwu: if I understand @mmlb  correctly?
```
if filter, ok := res.(string); ok {
		...
	} else if submenu, ok := res.(map[string]interface{}); ok {`
                ...
       } else {
                error
       }
```

http_handlers.go
```diff
@@ -102,6 +141,109 @@ func getMetadata(filter string) http.HandlerFunc {
...
+
+		sort.Strings(keys)
+
```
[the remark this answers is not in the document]
DavidHwu: do we run https://blog.golang.org/gofmt to normalize?

http_handlers.go
DavidHwu: Please consider handling all these use cases as well
```
if filter, ok := item.(string); ok { // if is an actual filter
			result = fmt.Sprint(base, filter)
		} else if subfilters, ok := item.(map[string]interface{}); ok { // if is another map of filters
			base = fmt.Sprint(base, subfilters["_base"].(string))
			result = subfilters
		} else {
			return nil, errors.New("invalid metadata item")
```

http_handlers.go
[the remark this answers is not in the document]
mmlb: yep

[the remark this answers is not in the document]
mmlb: if `res` is neither a string or map[string]interface{} what happens? Currently nothing, so it just keeps on chugging. We should be more defensive than that.

http_handlers.go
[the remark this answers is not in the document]
mmlb: yep, its part of ci. gofmt/goimports doesn't remove empty lines in function bodies though.

http_handlers.go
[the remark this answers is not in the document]
kdeng3849: Oh, I see now. Do I log the error and return internal server error then?

http_handlers.go
kdeng3849: Hmm is there a specific case you're thinking of? I can't think of anything I'm missing here https://github.com/tinkerbell/hegel/blob/2e1275844da2e97db4c4513bf3219f83b2458db2/http_handlers_test.go#L513

The first two test cases test when `item` is a string, the third one tests if `item`, the last three tests the error cases.

http_handlers.go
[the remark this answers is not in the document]
kdeng3849: fixed: https://github.com/tinkerbell/hegel/commit/b937ec7c49823ea9b4e8973b02c9ec3b06e7873e#diff-417a3d9f2a96ef7dc446d080e58f03b1R205

http_handlers.go
kdeng3849: fixed: https://github.com/tinkerbell/hegel/pull/22/commits/af31a740d81ab418184b58714a13d1118e059c87#diff-417a3d9f2a96ef7dc446d080e58f03b1R186

mmlb:
@DavidHwu you mentioned:
```
DavidHwu 5 days ago Member

IMO I am use to constructing the path dynamically and do not find that hard to read.
I see this as a fine/easier way to service and have the flexibility.
```
I'm not a fan of too much dynamic stuff and the `string/`map[string]interface{}` nature of the filter code wasn't sitting right with me. @kdeng3849 and I DM'd and I think we landed on something that will cut down on a lot of the complexity from this PR. A new PR will be opened so we can have a look at it separately.

http_handlers_test.go
```diff
@@ -171,63 +175,108 @@ func TestRegisterEndpoints(t *testing.T) {
...
+	os.Setenv("DATA_MODEL_VERSION", "1")
```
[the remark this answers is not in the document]
mmlb: like done here: https://github.com/tinkerbell/boots/blob/master/packet/models_test.go#L21-L23

mmlb:
this can be closed right @kdeng3849 ?

kdeng3849:
> this can be closed right @kdeng3849 ?

Yeah

PR #23, opened 2020-08-11
add subscription http endpoints

mikemrm:
## Description

Adds http endpoints to list or get subscribers. This allows for tooling to detect whether a server is currently connected to hegel to determine how it should proceed.

## Example Output

```
$ curl -s localhost:50061/subscriptions | jq '.'
{
  "abcdefab-cdef-abcd-efab-abcdefabcdef": {
    "id": "abcdefab-cdef-abcd-efab-abcdefabcdef",
    "ip": "127.0.0.1",
    "init_duration": 2000004,
    "started_at": "2020-08-11T20:55:04.56156941Z"
  }
}

$ curl -s localhost:50061/subscriptions/abcdefab-cdef-abcd-efab-abcdefabcdef | jq '.'
{
  "id": "abcdefab-cdef-abcd-efab-abcdefabcdef",
  "ip": "127.0.0.1",
  "init_duration": 2000004,
  "started_at": "2020-08-11T20:55:04.56156941Z"
}

$ curl -s localhost:50061/subscriptions/abcdefab-cdef-abcd-efab-abcdefabcdee | jq '.'
{
  "error": {
    "comment": "item not found",
    "error": "abcdefab-cdef-abcd-efab-abcdefabcdee not found"
  }
}
```


mikemrm:
@kdeng3849 tagging you on here for visibility. I've had to make quite a few changes in order to make hegel backwards compatible with osie/osie preinstall. Issues were primarily with metadata instance, and storage attributes.

main.go
```diff
@@ -84,15 +82,15 @@ func (hg hardwareGetterTinkerbell) Watch(ctx context.Context, in getRequest, opt
 }
 
 var (
-	facility = flag.String("facility", envString("HEGEL_FACILITY", "onprem"),
+	facility = flag.String("facility", env.Get("HEGEL_FACILITY", "onprem"),
 		"The facility we are running in (mostly to connect to cacher)")
-	tlsCertPath = flag.String("tls_cert", envString("HEGEL_TLS_CERT", ""),
+	tlsCertPath = flag.String("tls_cert", env.Get("HEGEL_TLS_CERT", ""),
```
mmlb, line 99: Can just be `env.Get("HEGEL_TLS_CERT")`

grpc_server.go
```diff
@@ -331,8 +337,36 @@ func (s *server) Subscribe(in *hegel.SubscribeRequest, stream hegel.Hegel_Subscr
...
+		Id:           id,
+		Ip:           ip,
+		StartedAt:    startedAt,
+		InitDuration: time.Since(startedAt),
+		cancel:       cancel,
+		updateChan:   make(chan []byte, 1),
+	}
+	s.subLock.Lock()
+	s.subscriptions[id] = sub
+	s.subLock.Unlock()
```
mmlb, line 311: suffers from TOCTOU. You're assuming after RUnlock the goroutine isn't being scheduled out and somehow a sub for this id is duplicated.
Instead you can do:
```go
	sub = &subscription{
		Id:           id,
		Ip:           ip,
		StartedAt:    startedAt,
		InitDuration: time.Since(startedAt),
		cancel:       cancel,
		updateChan:   make(chan []byte, 1),
	}
	s.subLock.Lock()
	old := s.subscriptions[id]
	s.subscriptions[id] = sub
	s.subLock.Unlock()

	// Disconnect previous client if a client is already connected for this hardware id
	if old != nil {
		old.cancel()
	}
```

grpc_server.go
```diff
@@ -331,8 +337,36 @@ func (s *server) Subscribe(in *hegel.SubscribeRequest, stream hegel.Hegel_Subscr
...
+
+	defer func() {
+		s.subLock.Lock()
+		defer s.subLock.Unlock()
+		if cSub, ok := s.subscriptions[id]; ok && cSub.updateChan == sub.updateChan {
```
mmlb, line 316: Don't need the `ok` check I think since for the `==` to be true then `ok` will also be true. If ok is false, there's no way `cSub.updateChan == sub.updateChan` because cSub.updateChan will be nil and `sub.updateChan` is never nil at this point.

mmlb: Also deserves some sort of comment that this is a guard against deleting a sub that has already been deleted because of a reconnection attempt or similar.

http_handlers.go
```diff
@@ -209,3 +210,56 @@ func getIPFromRequest(r *http.Request) string {
 	}
 	return IPAddress
 }
+
+func writeJson(w http.ResponseWriter, status int, data interface{}) error {
```
mmlb, line 214: should be `writeJSON`

http_handlers.go
```diff
@@ -209,3 +210,56 @@ func getIPFromRequest(r *http.Request) string {
...
+	var body []byte
+	body, err := json.Marshal(data)
+	if err != nil {
+		if status < 400 {
+			return jsonError(w, http.StatusInternalServerError, err, "Marshalling response")
```
mmlb, line 219: should be lower cased `marshalling`

http_handlers.go
```diff
@@ -209,3 +210,56 @@ func getIPFromRequest(r *http.Request) string {
...
+		} else {
+			logger.Error(err, "Failed to marshal error")
```
mmlb, line 221: `F` -> `f`

http_handlers.go
```diff
@@ -209,3 +210,56 @@ func getIPFromRequest(r *http.Request) string {
...
+			body = []byte(`{"error", {"comment": "Failed to marshal error"}}`)
```
mmlb, line 222: not a fan of error messages in body, its a bit annoying if you ask me. Do we need/actually use it?

main.go
```diff
@@ -31,6 +31,8 @@ import (
 type server struct {
 	log            log.Logger
 	hardwareClient hardwareGetter
+	subscriptions  map[string]*subscription
+	subLock        sync.RWMutex
```
mmlb, line 36: mutexes should show up above the thing(s) its protects, and should be separated by a new line from other struct fields.

main.go
```diff
@@ -41,6 +43,15 @@ type hardwareGetter interface {
 type getRequest interface{}
 type hardware interface{}
 
+type subscription struct {
+	Id           string        `json:"id"`
```
mmlb, line 48: `ID` please

main.go
```diff
@@ -41,6 +43,15 @@ type hardwareGetter interface {
...
+	Ip           string        `json:"ip"`
```
mmlb, line 49: `IP` please

grpc_server.go
```diff
@@ -47,18 +47,19 @@ type exportedHardwareCacher struct {
...
-	Rescue   bool   `json:"rescue"`
+	ID       string `json:"id,omitempty"`
+	State    string `json:"state,omitempty"`
+	Hostname string `json:"hostname,omitempty"`
+	AllowPXE bool   `json:"allow_pxe,omitempty"`
+	Rescue   bool   `json:"rescue,omitempty"`
 
-	OS       operatingSystem `json:"operating_system_version"`
-	UserData string          `json:"userdata,omitempty"`
+	IpAddresses []map[string]interface{} `json:"ip_addresses,omitempty"`
```
mmlb, line 53: `IPAddresses` please

grpc_server_test.go
```diff
@@ -119,14 +120,15 @@ func TestGetByIPTinkerbell(t *testing.T) {
 		if hw.Metadata.BondingMode != test.bondingMode {
 			t.Fatalf("unexpected bonding mode, want: %v, got: %v\n", test.bondingMode, hw.Metadata.BondingMode)
 		}
+		fmt.Println("before")
```
mmlb, line 123: remove?

main.go
mikemrm: good point

grpc_server.go
mikemrm: fair point, much cleaner as well

grpc_server.go
mikemrm: `cSub` could be `null` if the id doesn't already exist. But will add a comment

http_handlers.go
mikemrm: no, but it helps with troubleshooting an issue. This particular error is a catch all.

grpc_server.go
mmlb: right, but `sub` isn't null so they don't `==` so you wont delete

mmlb:
> This commit message is actually missing the most important part, that the suffix is very important to tools we run that uses this data, so needs to stay (sgdisk)

@mikemrm ^

grpc_server.go
mikemrm: on a normal disconnect, cSub and sub could be equal. As another connection would not have come in and overwrite the `s.subscriptions[id]`

mikemrm:
Sorry, missed that. I've added more details about the sgdisk in the comment.

PR #25, opened 2020-08-13
ENG-7133 add EC2 endpoint (v2)

kdeng3849:
## Description

Adds the EC2 endpoint `/2009-04-04`.
This is v2 of https://github.com/tinkerbell/hegel/pull/22.

## Why is this needed

There has been an effort to have Hegel succeed Kant, and one of the endpoints that Kant has (and Hegel doesn't) is the `/2009-04-04` endpoint. 

Fixes: #

## How Has This Been Tested?

- Unit & integration tests
- Manual test with tinkerbell setup

No migration steps needed.

## Checklist:

I have:

- [x] added unit or e2e tests


http_handlers_test.go
```diff
@@ -133,28 +154,33 @@ func TestGetMetadataTinkerbellKant(t *testing.T) {
...
+	dataModelVersion := os.Getenv("DATA_MODEL_VERSION")
+	defer os.Setenv("DATA_MODEL_VERSION", dataModelVersion)
 	os.Setenv("DATA_MODEL_VERSION", "1")
 
 	for name, test := range registerEndpointTests {
 		t.Log(name)
 		hegelServer.hardwareClient = hardwareGetterMock{test.json}
 
+		customEndpoints := os.Getenv("CUSTOM_ENDPOINTS")
+		defer os.Setenv("CUSTOM_ENDPOINTS", customEndpoints)
```
mmlb, line 166: This should be done outside the for loop.

grpc_server.go
```diff
@@ -196,16 +197,20 @@ func filterMetadata(hw []byte, filter string) ([]byte, error) {
...
+			if err != nil {
+				return nil, err
+			}
+			result.Write(vByte) // append v to result
+			result.WriteRune('\n')
+		}
 	}
-	return nil, nil
+
+	return bytes.TrimSuffix(result.Bytes(), []byte("\n")), nil
```
mmlb, line 213: hmm how does this look to you instead?
```go
if v == nil {
    continue
}

switch vv := v.(type) {
case error:
    return nil, errors.Wrap(vv, ...)
case string:
    result.WriteString(vv)
default:
	vByte, err := json.Marshal(v)
	if err != nil {
		return nil, errors.Wrap(err, ...)
	}
	result.Write(vv)
}
result.WriteRune('\n')
```
I like it because it keeps all the type checking in one block. Either way the errors returned need to be `errors.Wrap`ped.

http_handlers.go
```diff
@@ -82,26 +115,87 @@ func getMetadata(filter string) http.HandlerFunc {
...
+	}
+}
+
+func ec2Handler(w http.ResponseWriter, r *http.Request) {
+	if r.Method != "GET" {
+		w.WriteHeader(http.StatusMethodNotAllowed)
+		return
+	}
+
+	logger.Debug("Calling ec2Handler ")
```
mmlb, line 144: `C` -> `c`

grpc_server.go
```diff
@@ -415,20 +420,21 @@ func getByIP(ctx context.Context, s *server, userIP string) ([]byte, error) {
...
-		if resp == nil {
-			return nil, errors.New("could not find hardware")
-		}
-
 		hw, err = json.Marshal(util.HardwareWrapper{Hardware: resp.(*tink.Hardware)})
 		if err != nil {
 			return nil, errors.New("could not marshal hardware")
 		}
+
+		if string(hw) == "{}" { // workaround for resp is nil but resp != nil
```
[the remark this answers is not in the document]
kdeng3849: I wasn't sure how to concisely explain the problem where `resp` is `nil` (when printed during debug), but when I did `resp != nil` it's false

grpc_server_test.go
```diff
@@ -278,8 +297,75 @@ var tinkerbellGrpcTests = map[string]struct {
...
-		id:     "363115b0-f03d-4ce5-9a15-5514193d131a",
-		remote: "192.168.1.5",
-		json:   tinkerbellNoMetadata,
+		id:   "363115b0-f03d-4ce5-9a15-5514193d131a",
+		json: tinkerbellNoMetadata,
+	},
+}
+
+// test cases for TestFilterMetadata
+var tinkerbellFilterMetadataTests = map[string]struct {
```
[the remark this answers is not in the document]
kdeng3849: I'm not sure, I guess I wanted to put the focus of this test on just the filters and their responses. but this ^ does make it more dynamic

grpc_server_test.go
```diff
@@ -142,9 +145,27 @@ func TestGetByIPTinkerbell(t *testing.T) {
 	}
 }
 
+func TestFilterMetadata(t *testing.T) {
+	os.Setenv("DATA_MODEL_VERSION", "1")
```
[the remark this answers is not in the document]
kdeng3849: oops, actually not sure why that's even there, removed it.

grpc_server.go
kdeng3849: this looks so much better honestly

PR #27, opened 2020-08-19
ENG-8083 trusted proxies

kdeng3849:
## Description

Handles the case when hegel is running behind a reverse proxy.  
This was already handled for the grpc endpoints, so this pr is specifically for the http endpoints.

## Why is this needed

Hegel looks up the remote user ip from the request, then passes that ip to retrieve the associated piece of hardware. If hegel is behind a proxy, then the remote user ip will be the proxy's ip (and hegel won't be able to get the correct piece of hardware).

Fixes: #

## How Has This Been Tested?

- integration tests
- manual test

No migration needed.

## Checklist:

I have:

- [x] added unit or e2e tests


gxff/gxff.go
```diff
@@ -121,7 +123,7 @@ func parseTrustedProxies() []string {
 // If allowedSubnets is nil and TRUSTED_PROXIES is empty then X-FORWARDED-FOR will be ignored (no proxy is trusted).
 func New(l log.Logger, allowedSubnets []string) (grpc.StreamServerInterceptor, grpc.UnaryServerInterceptor) {
```
mmlb, line 124: We should rename this package to just `xff`, and change `New` to `GRPCMiddleware` and leave it to the caller to pass in a trusted proxies or not.

gxff/gxff.go
```diff
@@ -13,6 +13,8 @@ import (
 	"google.golang.org/grpc/peer"
 )
 
+var TrustedProxies = ParseTrustedProxies()
```
mmlb, line 16: lets not do this

http_handlers.go
```diff
@@ -209,3 +212,21 @@ func getIPFromRequest(r *http.Request) string {
 	}
 	return IPAddress
 }
+
+func handleTrustedProxies(mux *http.ServeMux) http.Handler {
```
mmlb, line 216: lets pass in the list of trusted proxies here instead of pulling it out from gxff

main.go
```diff
@@ -12,6 +12,8 @@ import (
 	"sync"
 	"time"
 
+	"github.com/prometheus/client_golang/prometheus/promhttp"
+
```
mmlb: this import move doesn't look right.

main.go
```diff
@@ -249,25 +250,7 @@ func main() {
...
-
-	logger.With("port", *metricsPort).Info("Starting http server")
-	go func() {
-		err := http.ListenAndServe(fmt.Sprintf(":%d", *metricsPort), nil)
-		if err != nil {
-			logger.Error(err, "failed to serve http")
-			panic(err)
-		}
-	}()
+	ServeHTTP()
```
mmlb, line 251: we should call `ParseTrustedProxies` and pass it to `ServeHTTP` here and also use it to pass to `GRPCMiddlewares` above.

http_server_test.go
```diff
@@ -11,9 +11,54 @@ import (
 	"strings"
 	"testing"
 
+	"github.com/packethost/hegel/gxff"
+
```
mmlb, line 15: should not have an empty line here

http_handlers.go
```diff
@@ -11,6 +11,9 @@ import (
 	"strings"
 	"time"
 
+	"github.com/packethost/hegel/gxff"
+	"github.com/sebest/xff"
+
```
mmlb, line 16: should remove the empty line

gxff/gxff.go
kdeng3849: As in, don't have a global variable? Can you elaborate on this?

main.go
kdeng3849: What do you mean by "move"? As in, this should stay in `main.go` instead of in `http_server.go` (where it was moved to)?

gxff/gxff.go
mmlb: yep lets not have a global / exported variable

xff/xff.go
```diff
@@ -148,3 +147,21 @@ func New(l log.Logger, allowedSubnets []string) (grpc.StreamServerInterceptor, g
 	}
 	return streamer, unaryer
 }
+
+func XFFHandler(l log.Logger, mux *http.ServeMux, allowedSubnets []string) http.Handler {
```
mmlb, line 151: I think maybe HTTPHandler makes more sense.

xff/xff.go
```diff
@@ -148,3 +147,21 @@ func New(l log.Logger, allowedSubnets []string) (grpc.StreamServerInterceptor, g
...
+		xffmw, err := xff.New(xff.Options{
+			AllowedSubnets: allowedSubnets,
+		})
+		if err != nil {
+			l.Fatal(err, "error creating a new xff handler")
+		}
+
+		handler = xffmw.Handler(mux)
+	} else {
+		handler = mux
```
mmlb, line 163: I think it would be nice to handle `mux` being nil and then using `http.DefaultServeMux`

xff/xff.go
```diff
@@ -119,18 +121,15 @@ func ParseTrustedProxies() []string {
 // New returns a set of grpc interceptors that will replace peer.IP with X-FORWARDED-FOR value if the peer ip within a subnet in allowedSubnets
 // If allowedSubnets is nil it will look for subents in the TRUSTED_PROXIES env var.
 // If allowedSubnets is nil and TRUSTED_PROXIES is empty then X-FORWARDED-FOR will be ignored (no proxy is trusted).
```
mmlb, line 123: Comment needs updating with the new name.

xff/xff.go
```diff
@@ -148,3 +147,21 @@ func New(l log.Logger, allowedSubnets []string) (grpc.StreamServerInterceptor, g
 	}
 	return streamer, unaryer
 }
+
```
mmlb, line 150: Needs a func comment

xff/xff.go
kdeng3849: 🤦‍♀️ thanks for catching that

xff/xff.go
```diff
@@ -118,8 +118,8 @@ func ParseTrustedProxies() []string {
 	return result
 }
 
-// New returns a set of grpc interceptors that will replace peer.IP with X-FORWARDED-FOR value if the peer ip within a subnet in allowedSubnets
-// If allowedSubnets is nil it will look for subents in the TRUSTED_PROXIES env var.
+// GRPCMiddlewares returns a set of grpc interceptors that will replace peer.IP with X-FORWARDED-FOR value if the peer ip within a subnet in allowedSubnets
```
mmlb, line 121: Oh I just noticed some grammatical issue here from before, can you fix though? Maybe to read like:
> if the peer IP is within one of the subents in allowedSubnets

kdeng3849: fixed: https://github.com/tinkerbell/hegel/pull/27/commits/884567ec2be5bdd2c541817fee72ec711a33fabb#diff-48a11bae66acafe529b7385a85a6a9e9R121

PR #29, opened 2020-08-26
Maintainers for hegel

gianarb:
We are in the process of assigning maintainers for all the projects.

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,0 +1,6 @@
+# These owners will be the default owners for everything in the repo. They
+# will be requested for review when someone opens a pull request.
+# We call those people maintainers as well
+
+*   @displague
+
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,2 +1,0 @@
-# Documentation https://help.github.com/articles/about-codeowners/
-*       @tinkerbell/site-eng


gianarb:
Looking at the contributions graph it will be nice if we can add @kdeng3849 as a maintainer as well :)

gianarb:
I just noticed that there is already a codeowners in .github. I would like to move it to the root of the project for visibility. I presume the community will benefit from having names close to repositories.

CODEOWNERS
```diff
@@ -0,0 +1,6 @@
+# These owners will be the default owners for everything in the repo. They
+# will be requested for review when someone opens a pull request.
+# We call those people maintainers as well
+
+*   @displague
```
mmlb, line 5: @tinkerbell/site-eng needs to be on here too please.

displague: I've been chatting with @mmlb, @kdeng3849, and @rainleander about this process.
Here are some of the points that are relevant to this PR:
* OWNERS.md should be used to define what it means to be an owner
  * This can / should link out to some common tinkerbell org-wide definition of ownership
  * "owner" and "maintainer" should be distinct definitions
    * What should an owner do? What should they not do (when do they defer to the community or proposals)
    * What expectations do we have for maintainers. 
    * Owner may be more of a shepherd, able to call the vet when needed, but otherwise feed and care for the animals, while a maintainer may be more of a vet and breeder. You've heard of clusters as cattle, how about code as sheep? 
* We may benefit from project-specific teams: `@tinkerbell/hegel`, `@tinkerbell/hegel-eng`? (I just made these names up, no idea if they are fitting. maybe `-owners` and `-maintainers` is better?)). The point is that each project may have maintainers that are not actively dealing with the code itself, but provide insight into documentation, branding/marketing, and an SRE perspective on processes and effects that a PR could affect on users.

displague: https://github.com/tinkerbell/hegel/pull/29/files#diff-5daf978e60e03b3975b485d5f1b449dcR1-R3 could be the start of OWNERS.md

We can add more details about the community (or better yet, link to it).

I'm ok with making all of these changes later once we get the basic/initial owners in place.

displague: Thanks for opening this PR for me @gianarb, so much better than nominating myself in a PR.

mmlb: >     * OWNERS.md should be used to define what it means to be an owner

This should be done before this PR is merged.

>       * This can / should link out to some common tinkerbell org-wide definition of ownership
>       * "owner" and "maintainer" should be distinct definitions

I nominate .github as this central place.

displague: I opened https://github.com/tinkerbell/hegel/pull/30 to add site-eng back into this branch and link to a .github/OWNERS.md file (tinkerbell/.github#4).

gianarb:
I don't like the idea to have `.md` file for that. It does not sound the right way to have an easy to automate approach.

Other than not having the right group assigned, somebody can tell me why CODEOWNERS does not work? GitHub uses that file to create automation, it is easy to parse (it means that we can create our own one as well). The community understands it pretty well because GitHub did that.

Now we have to take care of two files, codeowners and owners 👎 

---

Let me try to explain where this PR comes from because obviously it was a way to trigger a discussion and not a random action coming from nowhere :)

We need to create a link between repositories and real people. `site-eng` means everything and nothing to me, it is a group of superhero that can to everything and nothing :) Obviously I am joking but that's how I see it and the general plan should be as @displague described to split it out in more realistic and easy to figure out groups of people. I think we are not that big to require a `<repo>-admin` with admin permissions and `repo-<maintainers>` with maintainers permission for each repo. I think we should have a very small `admin` group kind of people that will have a look at the overall organization and a `<repo>-maintainers` for each repo.

I am saying we do not need a big `admin` group because my goal is to extend the [tinkerbell/org](https://github.com/tinkerbell/org) repo to act as a "GitHub as code" repository that the community can use to assign maintainers or ask for new repo. The `admin` will be the reviewer and owners of the org repo.

How does it sound? :) Let's get back on track. As a first step can we have a CODEOWNERS file that contains `site-eng` AND 2-3 name of people that we can reference to as a maintainer? Second step will be to expand the `org` repo and make a proper set of groups and so on.

Thanks

displague:
_It looks like I'm yelling with the filename, so I'm starting with this disclaimer instead of the filename to assure you that I'm not._ 😄 

OWNERS.md, in this case, is not a file that would need to be edited very often.  It does not track people.  OWNERS.md is a description of governance.  Maybe the name is confusing?

With the PRs that are up now, OWNERS.md would only live in `tinkerbell/.github`.

Each repo's CODEOWNERS file would link to the one Tinkerbell OWNERS.md file as an explanation of what the CODEOWNERS do and how they are appointed (and eventually more about how new people can become owners).

Would you feel differently if instead of being `tinkerbell/.github/OWNERS.md` it was a webpage at `tinkerbell.org/community#maintainers` (that's effectively what this is trying to do).

I don't have a strong opinion about groups-only, user-only, or groups-and-users format for CODEOWNERS. 
I'm happy with either choice so long as the CODEOWNERS includes the right people to keep the project healthy and act in the users' best interest.

gianarb:
Are you planning to use only one? OWNERS.md of CODEOWNERS? I have two desires:

1. I would like to have only one place where maintainers are listed. I would like for us to avoid guessing how many time a file will be read or written.
2. I would like to have that file easy to parse.

If we are looking for a place to write how a maintainer should act for the Tinkerbell community I am happy to see it written in the OWNERS.md if you like that. But we should remove the list of names from there because that file is not easy to parse and it duplicates informations

gianarb:
Will close in favor of #30 #31

PR #36, opened 2020-09-17
Update README.md

rainleander:
Signed-off-by: K Rain Leander <rainleander@users.noreply.github.com>

## Description

Our repositories should be the example from which adjacent, competing, projects look for inspiration.

Each repository should not look entirely different from other repositories in the ecosystem, having a different layout, a different testing model, or a different logging model, for example, without reason or recommendation from the subject matter experts from the community.

We should share our improvements with each ecosystem while seeking and respecting the feedback of these communities.

Whether or not strict guidelines have been provided for the project type, our repositories should ensure that the same components are offered across the board. How these components are provided may vary, based on the conventions of the project type. GitHub provides general guidance on this which they have integrated into their user experience.

## Why is this needed

Fixes: https://github.com/tinkerbell/hegel/issues/38

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,7 @@
 [![Build Status](https://cloud.drone.io/api/badges/tinkerbell/hegel/status.svg)](https://cloud.drone.io/tinkerbell/hegel)
+![](https://img.shields.io/badge/Stability-Experimental-red.svg)
+
+This repository is [Experimental](https://github.com/packethost/standards/blob/master/experimental-statement.md) meaning that it's based on untested ideas or techniques and not yet established or finalized or involves a radically new and innovative style! This means that support is best effort (at best!) and we strongly encourage you to NOT use this in production.
 
 ### Hegel
 


gianarb (approved):

README.md
```diff
@@ -1,4 +1,7 @@
 [![Build Status](https://cloud.drone.io/api/badges/tinkerbell/hegel/status.svg)](https://cloud.drone.io/tinkerbell/hegel)
+![](https://img.shields.io/badge/Stability-Experimental-red.svg)
+
+This repository is [Experimental](https://github.com/packethost/standards/blob/master/experimental-statement.md) meaning that it's based on untested ideas or techniques and not yet established or finalized or involves a radically new and innovative style! This means that support is best effort (at best!) and we strongly encourage you to NOT use this in production.
```
displague, line 4: The Tinkerbell community should have their own patterns for this, which can certainly be modeled on the Packet model. I imagine it will quickly evolve into its own thing.

Put succinctly, we should put the relevant https://github.com/packethost/standards/ bits in https://github.com/tinkerbell/.github and link there within Tinkerbell projects.

Thoughts, @gianarb @rainleander  ?

displague (approved):

README.md
rainleander: Definitely. But for the short term, this will work, eh?

mmlb (dismissed):

mmlb (approved):

mmlb:
Rebased on master and squashed the commits, as its really a fixup anyway.

PR #37, opened 2020-09-17
Create SUPPORT.md

rainleander:
Signed-off-by: K Rain Leander <rainleander@users.noreply.github.com>

## Description

Our repositories should be the example from which adjacent, competing, projects look for inspiration.

Each repository should not look entirely different from other repositories in the ecosystem, having a different layout, a different testing model, or a different logging model, for example, without reason or recommendation from the subject matter experts from the community.

We should share our improvements with each ecosystem while seeking and respecting the feedback of these communities.

Whether or not strict guidelines have been provided for the project type, our repositories should ensure that the same components are offered across the board. How these components are provided may vary, based on the conventions of the project type. GitHub provides general guidance on this which they have integrated into their user experience.

## Why is this needed

Fixes: https://github.com/tinkerbell/hegel/issues/31

### DIFF
--- a/SUPPORT.md
+++ b/SUPPORT.md
@@ -1,0 +1,1 @@
+If you require support, please subscribe to the [Packet Community Slack](https://slack.packet.com/) and join the #tinkerbell channel or post an issue within this repository.


gianarb (approved):

displague (approved):

mmlb (dismissed):

mmlb:
Rebased

mmlb (approved):

PR #41, opened 2020-10-05
Moved to Github Actions from drone CI

parauliya:
Signed-off-by: parauliya <aman@infracloud.io>

## Description

This chnage will move the CI from drone to Github Actions

## Why is this needed

Fixes: #39 

## How Has This Been Tested?
All checks are being passed while raising a PR

No impact.

## Checklist:

I have:

- [ ] added unit or e2e tests


.github/workflows/ci.yaml
```diff
@@ -0,0 +1,65 @@
...
+    - name: Build binaries
+      run: go build
+    - name: Upload hegel binary
+      uses: actions/upload-artifact@v2
+      with:
+        name: hegel
+        path: ./hegel
+  docker-images:
+    env:
+      CGO_ENABLED: 0
```
mmlb, line 42: this should be done in the validation part right? This isn't really doing anything here.

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,65 @@
...
+        version: v1.30
+        args: --skip-dirs cmd/packetflow
+    - name: go test
+      run: go test -v ./...    
+    - name: go test coverage
+      run: go test -coverprofile=coverage.txt ./...
+    - name: upload codecov
+      run: bash <(curl -s https://codecov.io/bash)
+    - name: Build binaries
+      run: go build
```
mmlb, line 34: Hmm is there an alpine builder? This is going to change hegel from building in alpine/musl to building in ubuntu/glibc. This is probably fine though since I don't think we have anything that will accidentally do a non-static build.

mmlb (changes requested):

.github/workflows/ci.yaml
parauliya: Yes, you are right. This will be removed.

gianarb (dismissed):
It looks very good to me! Let's wait for @mmlb because he knows a bit better Hegel, but I can't see reasons to hold this further.

Cbkhare (dismissed):

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,70 @@
...
+      CGO_ENABLED: 0
+    steps:
+    - name: Checkout code
+      uses: actions/checkout@v2
+    - name: Install Go
+      uses: actions/setup-go@v2
+      with:
+        go-version: '1.14.6'
+    - name: go fmt
+      run: go get golang.org/x/tools/cmd/goimports; goimports -d . | (! grep .)
```
mmlb, line 19: this should be `&&`. I think at one point there was a typo or something and the get failed but the command continued because of `;` and goimports failed but because there was no `pipefail` it didn't error and `! grep` worked out fine. Not in this repo I think, maybe another one.

mmlb (changes requested):

.github/workflows/ci.yaml
parauliya: Done

Cbkhare (approved):

mmlb (approved):
Pretty sure we're going to remove the `pull_request` bit pushing to quay since secrets aren't provided for PRs. But we can do that in a follow up PR.

gianarb (approved):

PR #45, opened 2020-10-21
Login to quay should only work if commit is in master

parauliya:
## Description

GitHub Action environment variables gets passed to PR only if they
starts from a branch that is local, so fork does not get a QUAY PASSOWRD
and USERNAME, this is to avoid security issue.

We should push images to Quay only if the commit is in master.


Cbkhare (approved):

gianarb (approved):

mmlb (approved):

gianarb:
This change is included #47 , if we can that merged, I will avoid a rebase :)

PR #49, opened 2020-11-05
Change default metadata endpoint handling

mmlb:
## Description

Change the handling the /metadata endpoint so the data returned is of the same content as the /meta-data EC2 style endpoint.

## Why is this needed

Hegel is meant as a replacement for what is used in EM Production today (Kant). The response from /metadata and /meta-data Kant returns is the same content. Its also arguably confusing to have 2 similarly named endpoints return very different data. This confusion is in part due to not having an Instance object yet expecting instance data _somewhere_. This should be cleaned up/made to make sense by https://github.com/tinkerbell/proposals/pull/25.

## How Has This Been Tested?

Unit tests still pass.

This is a bc-break. If any users were using the /metadata endpoint in its default configuration their workflow would be broken. I'd like to wait and see if thats the case before prematurely having any workarounds.


http_server_test.go
```diff
@@ -163,17 +158,18 @@ func TestGetMetadataTinkerbellKant(t *testing.T) {
 	defer os.Setenv("DATA_MODEL_VERSION", dataModelVersion)
 	os.Setenv("DATA_MODEL_VERSION", "1")
 
-	customEndpoints := os.Getenv("CUSTOM_ENDPOINTS")
-	defer os.Setenv("CUSTOM_ENDPOINTS", customEndpoints)
-	os.Setenv("CUSTOM_ENDPOINTS", `{"/metadata":".metadata.instance","/components":".metadata.components","/userdata":".metadata.userdata"}`)
-
+	customEndPoints := map[string]string{
```
kdeng3849, line 160: super nitpicky i know, but can you lower case the the p here (`customEndpoints`)?

http_server_test.go
```diff
@@ -554,18 +547,6 @@ var registerEndpointTests = map[string]struct {
...
-		status: 200,
-		json:   tinkerbellDataModel,
-	},
-	"custom endpoints invalid format (not a map)": {
-		customEndpoints: `"/userdata":".metadata.userdata"`,
-		url:             "/userdata",
-		status:          404,
-		error:           "error in parsing custom endpoints: invalid character ':' after top-level value",
-		json:            tinkerbellDataModel,
-	},
```
kdeng3849, line 564: what's the reasoning for removing these two cases?   
shouldn't we have a case that tests the default? and the second one is for covering the bases of all the possible invalid envvar formats

kdeng3849:
With this change, would it make sense that we add in a `CUSTOM_ENDPOINTS: '{"/metadata":".metadata"}'` in tink's docker-compose file? (I think James @splaspood also mentioned this earlier but I could be wrong.)

I understand the reasoning behind this change but I'm still a little e_e.  
But with this^ we can omit the envvar when we bring up hegel, but for the open source community who (I assume) mostly use the docker compose file, they can "default" to `"/metadata": ".metadata"`? What do you think?

mmlb:
Yeah that is a good option. I kind of would rather merge this as is though and see if there are any problems with it. I'm not sure if any workflows are even using hegel/metadata (https://deploy-preview-162--suspicious-jackson-6c9a73.netlify.app/blog/2020/10/provisioning-flatcar-container-linux-with-tinkerbell/ just embeds the ignition straight into the template). So I'd rather err on the side of /metadata is same as /meta-data unless we need to come up with a work around.

http_server_test.go
mmlb: :+1:

main.go
```diff
@@ -235,17 +235,22 @@ func main() {
...
-	}
-
+func parseCustomEndpoints(customEndpoints string) map[string]string {
 	endpoints := make(map[string]string)
 	err := json.Unmarshal([]byte(customEndpoints), &endpoints)
 	if err != nil {
-		return errors.Wrap(err, "error in parsing custom endpoints")
+		err = errors.Wrap(err, "error in parsing custom endpoints")
+		logger.Fatal(err)
+		panic(err)
```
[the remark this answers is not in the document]
mmlb: yep, thats left over from when I copy/pasted. Will fix.

http_server_test.go
mmlb: I had originally done the CUSTOM_ENDPOINTS parsing in `func main()` and so removed it. But I didn't want to affect codecoverage too much so put it behind the func we have now. But I forgot to bring this test back in. Will fix.

main.go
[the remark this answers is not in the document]
mmlb: fixed, PTAL

http_server_test.go
mmlb: fixed, PTAL

http_server_test.go
mmlb: fixed, PTAL. I left the default one out as thats done in `func main()` now, as it should be :D.

mmlb:
alrighty @kdeng3849 I think everything should all set now.

kdeng3849 (dismissed):

kdeng3849 (approved):

PR #51, opened 2020-12-01
Use operating_system_version vs operating_system | ENG-11034

splaspood:
## Description

Metadata, as passed into tink, stores operating system information under an operating_system_version key.   Hegel, currently, expects 'operating_system' instead when constructing the corresponding EC2 endpoints.

## Why is this needed

Since the data is actually stored, in tink, as 'operating_system_version' hegel needs to reference the data using the same key.   This is a bug.


splaspood:
this PR was made in error, closing.

PR #52, opened 2020-12-08
Hotfix - use DeprecatedWatch for now

gauravgahlot:
## Description

The PR fixes build issues in Hegel due to recent changes in Tink wrt to event-driven design.
For now, Hegel will continue using the old watch style with `DeprecatedWatch`.
Changes wrt to the new watch stream will be done in a separate PR. 

## Why is this needed

The PR fixes build issues in Hegel due to recent changes in Tink wrt to event-driven design.


kdeng3849 (approved):

grpc_server_test.go
```diff
@@ -139,8 +139,8 @@ func TestGetByIPTinkerbell(t *testing.T) {
 				t.Fatalf("unexpected filesystem mount format, want: %v, got: %v\n", test.filesystemFormat, hw.Metadata.Instance.Storage.Filesystems[0].Mount.Format)
 			}
 		}
-		if hw.Metadata.Instance.OperatingSystemVersion.OsSlug != test.osSlug {
-			t.Fatalf("unexpected os slug, want: %v, got: %v\n", test.osSlug, hw.Metadata.Instance.OperatingSystemVersion.OsSlug)
+		if hw.Metadata.Instance.OperatingSystem.Slug != test.osSlug {
```
kdeng3849, line 142: Maybe it's the `hw.Metadata.Instance` that is causing the nil pointer dereference

kdeng3849 (changes requested):

grpc_server_test.go
gauravgahlot: @kdeng3849 
I made the following corrections in metadata JSON:

```
operating_system_version   > operating_system
os_slug   > slug
```
Tests pass but I'm not certain if it is doing what's expected.

At the `matster` branch when I debug the test `hw.Metadata` fields are NIL. If we print the [ehw](https://github.com/tinkerbell/hegel/blob/master/grpc_server_test.go#L103) this is what we have and it doesn't contain metadata. Is it expected?
```
{"id":"363115b0-f03d-4ce5-9a15-5514193d131a","network":{"interfaces":[{"dhcp":{"arch":"x86_64","hostname":"server001","ip":{"address":"192.168.1.5","gateway":"192.168.1.1","netmask":"255.255.255.248"},"lease_time":86400,"mac":"ec:0d:9a:c0:01:0c"},"netboot":{"allow_pxe":true,"allow_workflow":true,"ipxe":{"contents":"#!ipxe","url":"http://url/menu.ipxe"},"osie":{"kernel":"vmlinuz-x86_64"}}}]}}
```

kdeng3849: Turns out it was returning on the first test case with empty metadata when it should really be `continue`. Can you change that for me (also update the comment)? Thanks
https://github.com/tinkerbell/hegel/blob/43d76c5eed1e630f27ae688a5c16e639894a446a/grpc_server_test.go#L113

gauravgahlot: Done. 👍🏼

grpc_server_test.go
```diff
@@ -110,8 +110,8 @@ func TestGetByIPTinkerbell(t *testing.T) {
 				hw.ID, test.id)
 		}
 
-		if reflect.DeepEqual(&hw.Metadata, &packet.Metadata{}) { // return if metadata is empty
-			return
+		if reflect.DeepEqual(&hw.Metadata, &packet.Metadata{}) { // continue if metadata is empty
```
Cbkhare, line 113: I think this `If` condition is not required. Are you keeping it for some future use?

gauravgahlot: It's required to have a mix of test cases where some hardware data may have metadata and others may not. If we remove this and the hardware data does not metadata, it will error out at line 116.

Cbkhare (approved):

PR #66, opened 2021-09-15
Add local-hostname to metadata

detiber:
## Description

Adds local-hostname to the ec2 style metadata output to allow for better consumption of metadata by cloud-init

## Why is this needed

When cloud-init attempts to set the hostname using the ec2 metadata source, it uses the value of local-hostname.

## How Has This Been Tested?

Ran existing test suite 

This should not affect existing users, other than those that are hitting issues with cloud-init not setting the hostname when using hegel as an ec2 metadata source.


jacobweinstock (approved):
/ltgm

displague (approved):

mmlb (approved):

PR #70, opened 2021-12-13
Update to latest github.com/packethost/pkg (OTel changes)

tstromberg:
The latest `packethost/pkg` fixes some OTel panic conditions. This PR was generated from:

`go get github.com/packethost/pkg && go mod tidy`

This PR may fix #69

PR #71, opened 2021-12-13
Update pkg and otel

tobert:
Updates the `pkg` package to the latest sha to fix the panic.

Updates `otel-init-go` to v0.0.4 to fix the otel compatibility thing.

Removes a span event that is redundant with those provided by gRPC autoinstrumentaiton.


tobert:
Replaces #70

tstromberg (approved):

Issue #69, opened 2021-12-13
otelgrpc.StreamServerInterceptor -> panic: runtime error: invalid memory address or nil pointer dereference


tstromberg:
Capturing a panic we saw in production using OTEL to record trace spans:

```
2021-12-13T20:25:49.691Z	INFO	grpc-server/grpc_server.go:197		{"service": "github.com/tinkerbell/hegel", "op": "subscribe", "ip": "#.#.#.#", "client": "#.#.#.#:57404"}
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
	panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x47e82d]

goroutine 219 [running]:
go.opentelemetry.io/otel/sdk/trace.(*span).End(0xc000305200, 0x0, 0x0, 0x0)
	/go/pkg/mod/go.opentelemetry.io/otel/sdk@v1.0.0-!r!c2/trace/span.go:274 +0x319
panic(0xb99f20, 0x11b8ef0)
	/usr/local/go/src/runtime/panic.go:969 +0x1b9
sync.(*RWMutex).Lock(0x0)
	/usr/local/go/src/sync/rwmutex.go:98 +0x2d
github.com/tinkerbell/hegel/grpc-server.(*Server).Subscribe(0xc0001b0f00, 0xc0006dc000, 0xd7b8c0, 0xc000313ec0, 0x0, 0x0)
	/usr/myapp/grpc-server/grpc_server.go:225 +0x885
github.com/tinkerbell/hegel/grpc/protos/hegel._Hegel_Subscribe_Handler(0xc07220, 0xc0001b0f00, 0xd792e0, 0xc000255fb0, 0x15, 0xc000395f60)
	/usr/myapp/grpc/protos/hegel/hegel.pb.go:311 +0x10b
go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc.StreamServerInterceptor.func1(0xc07220, 0xc0001b0f00, 0xd79220, 0xc000395e40, 0xc000395c60, 0xcb5c70, 0x0, 0x0)
	/go/pkg/mod/go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc@v0.22.0/interceptor.go:414 +0x6ce
[...]

PR #74, opened 2022-01-06
Fix Subscribe panic and add a test which would have caught it

nshalman:
## Description

This PR creates a new test for the Subscribe endpoint as well as fixing a current bug which causes panics when that RPC is used.

## Why is this needed

Fixes: #69

## How Has This Been Tested?

```console
~/Workspace/hegel % go test -run TestSubscribe -count 2000 ./grpc-server/...
ok  	github.com/tinkerbell/hegel/grpc-server	15.102s

~/Workspace/hegel % git revert --no-edit 88728c4
Auto-merging grpc-server/grpc_server.go
[nshalman/subscribe-fix-and-test 1d53e08] Revert "subLock needs to be initialized"
 Date: Fri Jan 7 13:14:24 2022 -0500
 1 file changed, 1 deletion(-)

~/Workspace/hegel % go test -run TestSubscribe ./grpc-server/...
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x107525e]

goroutine 104 [running]:
sync.(*RWMutex).Lock(0xc00031d080)
<snip>
```

## Checklist:

I have:

- [x] added unit or e2e tests


nshalman:
Thanks again to @mmlb for most of this test harness, and for the pointers for the further cleanup of the tests in https://github.com/tinkerbell/hegel/compare/669989512f74834d3a41cbcc271cf03cb152c630..f0ae94cd34199f0d1d1ce3c90e64dfb8ad9111a5

mmlb:
> Thanks again to @mmlb for most of this test harness, and for the pointers for the further cleanup of the tests in [`669989512f74834d3a41cbcc271cf03cb152c630..f0ae94cd34199f0d1d1ce3c90e64dfb8ad9111a5` (compare)](https://github.com/tinkerbell/hegel/compare/669989512f74834d3a41cbcc271cf03cb152c630..f0ae94cd34199f0d1d1ce3c90e64dfb8ad9111a5)

np. now that I look again the `WaitGroup` is probably unnecessary and I forgot to drop it

nshalman:
> now that I look again the `WaitGroup` is probably unnecessary and I forgot to drop it

Good catch. I will clean that up.

tobert (approved):
I read it over and think I understand what's happening and it looks fine to me. Thanks for adding this!

grpc-server/grpc_helpers_test.go
```diff
@@ -0,0 +1,232 @@
...
+// Logs the given message and marks the test as failed.
+func (zt *zapT) Errorf(format string, args ...interface{}) {
+	zt.mu.RLock()
+	if !zt.done {
+		zt.T.Logf(format, args...)
+	}
+	zt.mu.RUnlock()
+}
+
+// serveAndDial starts 2 grpc services.
```
mmlb, line 48: should be `startServersAndConnectClient`

grpc-server/subscribe_test.go
```diff
@@ -0,0 +1,114 @@
...
+		assert.NoError(t, err)
+		assert.NotNil(t, w)
+
+		hw, err := w.Recv()
+		assertGRPCError(t, err, terr)
+		assert.Nil(t, hw)
+	})
+
+	t.Run("good", func(t *testing.T) {
+		// id := uuid.Must(uuid.NewRandom()).String()
```
mmlb, line 46: should be deleted

grpc-server/subscribe_test.go
```diff
@@ -0,0 +1,114 @@
...
+		id := "bufconn"
+		data := map[string]string{}
+		value := fmt.Sprintf(`{"id": "%s", "ip": "%s"}`, id, id)
+		payload := fmt.Sprintf(`%s=%s`, id, value)
+		data[id] = value
```
mmlb, line 51: could be:
```suggestion
		value := fmt.Sprintf(`{"id": "%s", "ip": "%s"}`, id, id)
		payload := fmt.Sprintf(`%s=%s`, id, value)
		data := map[string]string{
			id: value,
		}
```
but not a big deal to change really

mmlb (changes requested):
just a few minor nits the only one I'd really like to see fixed is the comment on `startServersAndConnectClient`

grpc-server/grpc_helpers_test.go
nshalman: And the rest of the comment updated too. 😄

grpc-server/grpc_server.go
```diff
@@ -61,9 +61,11 @@ func NewServer(l log.Logger, hc hardware.Client) (*Server, error) {
 		}
 	}
 
+	var m sync.RWMutex
 	s := &Server{
 		log:            l,
 		hardwareClient: hc,
+		subLock:        &m,
```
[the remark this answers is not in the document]
nshalman: Done

grpc-server/grpc_helpers_test.go
nshalman: Done

grpc-server/subscribe_test.go
nshalman: Done

grpc-server/subscribe_test.go
nshalman: Done

grpc-server/subscribe_test.go
```diff
@@ -0,0 +1,114 @@
...
+		}()
+
+		hw := cacher.Hardware{}
+		assert.NoError(t, w.RecvMsg(&hw))
+		orig := hw.JSON
+		close(initialPushes)
+
+		// ok now that w.RecvMsg has processed at least one push we can run the actual test.
+		// We need to accept plain old $id as the value because the initialPush go routine
+		// could have snuck in a couple extra pushes while we get back from the w.RecvMsg call
```
nshalman, line 86: This comment should be updated too

mmlb (approved):
:shipit:

Issue #67, opened 2021-09-17
remove or fix hegelc


displague:
## Expected Behaviour

It is not clear what functionality hegelc adds to this project. 

## Current Behaviour
If this is a useful component, the default hostname should not be metadata.packet.net.

## Possible Solution

Remove hegelc or improve the README.md to explain the purpose and change the hostname to be configurable (environment variable, command line) with a default like 'hegel' or 'metadata' that will work well in docker-compose, mdns, or DNS domain search environments.

## Steps to Reproduce (for bugs)

1.
2.
3.
4.

## Your Environment

* Operating System and version (e.g. Linux, Windows, MacOS):

* How are you running Tinkerbell? Using Vagrant & VirtualBox, Vagrant & Libvirt, on Packet using Terraform, or give details:

* Link to your project or a code example to reproduce issue:

splaspood:
Excellent call out re metadata.packet.net.   I think a hegel 'client' is useful for development and we should probably enhance hegelc to support all the various grpc functions available in the hegel 'API'.

displague:
Created https://github.com/tinkerbell/hegel/pull/76 to address this

PR #76, opened 2022-01-11
Hegelc cleanup

displague:
## Description

* Moves `/hegelc` to `/cmd/hegelc`
* Adds environment variables (`HEGEL_SERVER`, `HEGEL_PORT`)
* Adds flags (`-server`, `-port`)
* changes default hostname to `metadata` from `metadata.packet.net`
* default error behavior is `Fatal` (exit code 1) rather than `Panic` (exit code 2)
* use `grpc.DialContext` instead of `grpc.Dial`
* fixes gocritic warnings about [exitAfterDefer](https://go-critic.com/overview.html#exitafterdefer)

## Why is this needed

Fixes: #67 

## How Has This Been Tested?

ran with `-help`, env variables set, flags set. Attempted invalid ports and hostnames and got appropriate errors.

The default server changes from metadata.packet.net to metadata.

## Checklist:

I have:

- [x] updated the documentation and/or roadmap (if required)
- [ ] added unit or e2e tests


mmlb:
Note that log.Panic panics and thus defered functions are run, but log.Fatal I think calls os.Exit and does not run defers. I think that's fine though and better for a cli than dumping a stack trace anyway.

cmd/hegelc/main.go
```diff
@@ -4,36 +4,65 @@ import (
...
+)
+
 func main() {
+	if envServer := os.Getenv(envVarServer); envServer != "" {
+		defaultServer = envServer
+	}
+	if envPort := os.Getenv(envVarPort); envPort != "" {
+		var err error
+		if defaultPort, err = strconv.Atoi(envPort); err != nil {
+			log.Panic(err)
```
mmlb, line 37: this should be log.Fatal too right?

mmlb, line 37: fyi: github.com/packethost/pkg/env has a bunch of nice funcs that do these conversions, you might want to use that, but not going to require it.

mmlb (changes requested):

cmd/hegelc/main.go
displague: The linter caught me and I changed more of this around.

mmlb (approved):
not a fan of basically empty main, but oh well not really a big deal either

Issue #62, opened 2021-07-06
Implement vendor-data paths


displague:
## Expected Behaviour

Users providing metadata may wish to include vendor-data in the metadata. Vendor data is like userdata, but it is only run on first boot and can be overridden by userdata. https://cloudinit.readthedocs.io/en/latest/topics/vendordata.html

This vendor data is currently not exposed through the EC2 compatibility paths.

https://github.com/tinkerbell/hegel/blob/9f5da0a80af990031230b0698d56606787b16eec/http-server/http_handlers.go#L26-L49

Tools like cloud-init know to look for vendor-data. https://github.com/canonical/cloud-init/blob/35aa9db6f8e2ba05d366776c0e8d97f52217e930/cloudinit/util.py

See also, https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instancedata-data-categories.html

## Current Behaviour

vendor-data is not exposed through the EC2 metadata.

## Possible Solution

Implement vendor-data, treating it as a URL that acts like userdata, supplied by a vendordata metadata key.

## Steps to Reproduce (for bugs)

1.
2.
3.
4.

## Your Environment

* Operating System and version (e.g. Linux, Windows, MacOS):

* How are you running Tinkerbell? Using Vagrant & VirtualBox, Vagrant & Libvirt, on Packet using Terraform, or give details:

* Link to your project or a code example to reproduce issue:

jacobweinstock:
Hey @displague, wanted to just check that this is still outstanding, yes?

chrisdoherty4:
Having discussed what Hegel's API is in the community meeting and with the development of the Hegel metadata API I'm closing this in favor of waiting for the Hegel metadata.

Also see #77

PR #77, opened 2022-01-12
add vendor-data EC2 style path (like user-data)

displague:
## Description

When "vendordata" is included in the hardware metadata, expose it
through the EC2 path structure.

Although vendor-data does not appear to be an EC2 exposed path, tools
like cloud-init make all of the paths listed in 2009-04-04 available to
scripts and cloud-init takes advantage of these paths, including
vendor-data.

## Why is this needed

Fixes: #62

## Checklist:

I have:

- [ ] added unit or e2e tests


displague:
I don't believe this has a hard dependency on https://github.com/tinkerbell/tink/pull/579, but it may be worth looking at first.

jacobweinstock:
Anyone more familiar with this code base able to review? @mmlb , @nshalman ?

chrisdoherty4:
@displague Vendor data isn't strictly part of the Equinix metadata standard. I'm not totally sure we should add it. 

Given we're developing the new Hegel metadata standard how would you feel about exposing vendor data in that instead?

chrisdoherty4:
Closing, see #62.

PR #81, opened 2022-04-07
Generalize the hardware client interface

chrisdoherty4:
### Context

To compliment the [Kubeification of Tinkerbell's back-end](https://github.com/tinkerbell/proposals/tree/main/proposals/0026) we need to provide a mechanism for supplying hardware data from the Kubernetes API Server in Hegel.

### Changes

The hardware client is used to interact with a Cacher service and the Tink service. It embodies gRPC specific types and an ambiguous All() call that are difficult to model for a Kubernetes data provider.

This change alters the interface to model more generic behavior and removes gRPC specific types paving the path for implementing a Kubernetes data provider client that can plug into existing business logic.

This change doesn't alter any service behavior.

### DIFF
--- a/hardware/client.go
+++ b/hardware/client.go
@@ -13,7 +13,6 @@ import (
 	tinkClient "github.com/tinkerbell/tink/client"
 	tpkg "github.com/tinkerbell/tink/pkg"
 	tink "github.com/tinkerbell/tink/protos/hardware"
-	"google.golang.org/grpc"
 )
 
 var (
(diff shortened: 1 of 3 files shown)


thebsdbox (dismissed):
I'm loving the push towards Kubernetes as a back end. Thanks

hardware/client.go
```diff
@@ -13,22 +13,26 @@ import (
...
-// Client acts as the messenger between Hegel and Cacher/Tink.
+// Client defines the behaviors for interacting with hardware data providers.
 type Client interface {
-	All(ctx context.Context, opts ...grpc.CallOption) (AllClient, error)
-	ByIP(ctx context.Context, id string, opts ...grpc.CallOption) (Hardware, error)
-	Watch(ctx context.Context, id string, opts ...grpc.CallOption) (Watcher, error)
-}
+	// IsHealth reports whether the client is connected and can retrieve hardware data from the data provider. Clients
+	// should report nil if healthy.
+	IsHealthy(ctx context.Context) error
```
mmlb, line 27: I'd expect Is* to return bool not error. If it were a bool we wouldn't need the comment explaining how to signal health either.

mmlb (changes requested):
One small nit, otherwise lgtm. Thanks!

hardware/client.go
chrisdoherty4: I had it that way originally and thought the semantics for 'unhealthy' might need inspecting by consumers to decide if they care or not. 

I'll change to bool as I prefer the simplicity of it.

chrisdoherty4: Adjusted. Feel free to unresolve if you have another thought.

mmlb (approved):

mmlb (approved):

PR #82, opened 2022-04-15
Add Kubernetes back-end client

chrisdoherty4:
## Description

Builds on #95 (rebased after PR was created)

This feature builds on #81 and realizes the Hegel portion of the [Kubernetes back-end proposal](https://github.com/tinkerbell/proposals/tree/main/proposals/0026).

Having generalized the client used to acquire metadata, this PR implements a Kubernetes client that leverages an internal client caching mechanism so as to reduce load on the Kubernetes API Server.

Additionally, I put some effort into reworking the Hegel Dockerfile so it could leverage layer caching for Go module dependencies and added an image build command to the Makefile.

The `Watch` API for the client is unimplemented and scheduled for a follow up. This is so we can get off the ground for machine provisioning. The `Watch` API is used for subscription capabilities so doesn't feature in metadata retrieval. 

## How Has This Been Tested?

Built and ran the image with the Kubernetes back-end enabled talking to a KinD cluster containing the necessary CRDs and a `Hardware` resource. 

None. Users enable the Kubernetes back-end by setting the `DATA_MODEL_VERSION=kubernetes` plus some additional env vars.

## Future work

- Implement the `Watch` API for consistency with other back-ends.


hardware/kubernetes.go
```diff
@@ -0,0 +1,132 @@
...
+}
+
+// WaitForCacheSync waits for the internal client cache to synchronize.
+func (k *KubernetesClient) WaitForCacheSync(ctx context.Context) bool {
+	return k.waitForCacheSync(ctx)
+}
+
+// IsHealthy always returns true.
+func (k *KubernetesClient) IsHealthy(context.Context) bool {
+	return true
```
chrisdoherty4, line 89: Suggestions on what to do here as its not true. Initially I thought a mutex protected `k.close` then set `k.close = nil` in the `Close()` func with a `return k.close != nil` for `IsHealthy()`. @micahhausler is there a better way in the manager to check if we're 'connected'?

hardware/client.go
```diff
@@ -71,24 +75,51 @@ type watcherTinkerbell struct {
...
 }
 
 // NewClient returns a new hardware Client, configured appropriately according to the mode (Cacher or Tink) Hegel is running in.
 func NewClient() (Client, error) {
 	var hg Client
 
 	switch dataModelVersion {
-	case "1":
+	case datamodel.Kubernetes:
+		kubeAPI := os.Getenv("HEGEL_KUBE_API")
```
micahhausler, line 91: I'm trying to standardize components on the same flags. Can we go with flags for `kubernetes-api` and `kubeconfig` and then ingest with viper here?

hardware/kubernetes.go
micahhausler: It doesn't look like controller-runtime's client exposes any health metrics.I think this is good enough

hardware/kubernetes.go
```diff
@@ -0,0 +1,132 @@
...
+		}
+		return nil, fmt.Errorf("multiple hardware with ip '%v': [%v]", ip, strings.Join(names, ", "))
+	}
+
+	return &hardware{hw.Items[0]}, nil
+}
+
+// Watch is unimplemented.
+func (k *KubernetesClient) Watch(context.Context, string) (Watcher, error) {
+	return nil, errors.New("kubernetes client: watch is unimplemented")
```
micahhausler, line 119: Why is watch not implemented here?

hardware/kubernetes.go
```diff
@@ -0,0 +1,132 @@
...
+}
+
+type hardware struct {
+	tinkv1alpha1.Hardware
+}
+
+func (h hardware) Export() ([]byte, error) {
+	return json.Marshal(h.Hardware.Spec)
```
micahhausler, line 127: What structure does Export expect here? Is the CRD spec going to be ingested correctly?

hardware/kubernetes_test.go
```diff
@@ -0,0 +1,30 @@
...
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"github.com/tinkerbell/hegel/hardware"
+	"k8s.io/client-go/tools/clientcmd"
+)
+
+func TestKubernetesClient(t *testing.T) {
+	config, err := clientcmd.BuildConfigFromFlags("", "/Users/cpd/.kube/config")
```
micahhausler, line 15: Can we use envtest instead of a hardcoded kubeconfig?

micahhausler (changes requested):
A few questions, and some cleanup.

hardware/kubernetes_test.go
chrisdoherty4: This is a local test I put together for ease. Its going to be replaced entirely with unit tests and e2e tests. 

The e2e tests should address the [export comment](https://github.com/tinkerbell/hegel/pull/82#discussion_r851402168).

hardware/kubernetes.go
chrisdoherty4: [See my other reply](https://github.com/tinkerbell/hegel/pull/82#discussion_r851406714). 

I think there are some inconsistencies. There will likely be a need to translate between the CRD spec and the export expectations but some fields match up already hence I was able to validate things "work".

hardware/kubernetes.go
chrisdoherty4: `Watch` is only used by the gRPC service. I've been told its used by Equinix Metal but the metadata capability doesn't need it, I don't think, hence I've left it out of this PR to be implemented in a follow up.

The changes in this PR should get us off the ground with respect to machine provisioning.

hardware/client.go
chrisdoherty4: Hegel doesn't support CLI flags at present hence I purposely avoided adding anything for this PR. 

I'd like to address it separately because I think it'd be worth introducing CLI arguments holistically; does that work for you?

.gitattributes
```diff
@@ -0,0 +1 @@
+go.sum -diff
```
chrisdoherty4, line 1: This won't take effect until merged but it'll exclude `go.sum` from the loc on PRs.

hardware/kubernetes_client.go
```diff
@@ -0,0 +1,180 @@
...
+		Namespace: namespace,
+	}, nil
+}
+
+type hardware struct {
+	tinkv1alpha1.Hardware
+}
+
+func (h hardware) Export() ([]byte, error) {
+	return json.Marshal(h.Hardware.Spec)
```
chrisdoherty4, line 175: Unit test

hardware/kubernetes.go
chrisdoherty4: I've encoded the minimal expectations into a new export construct to be safe (there were some problems anyhow so this fixes them). 

It would be useful to move away entirely from this approach but it likely means breaking compatibility with the custom metadata rules that can be added via an env var on startup. I suspect others are using that capability which makes it difficult to detach from the jq parsing approach and head toward something that's more clearly defined.

.golangci.yml
```diff
@@ -31,6 +31,9 @@ linters-settings:
   errorlint:
     # these are still common in Go: for instance, exit errors.
     asserts: false
+    # Forcing %w in error wrapping forces authors to make errors part of their package APIs.
+    # Also see Hyrums Law.
+    errorf: false
```
displague, line 36: This sounds like the default. Any reason to set it explicitly?
https://github.com/polyfloyd/go-errorlint/blob/master/README.md#examples

go.mod
```diff
@@ -26,10 +25,12 @@ require (
...
 	golang.org/x/net v0.0.0-20220105145211-5b0dc2dfae98 // indirect
-	golang.org/x/sys v0.0.0-20211216021012-1d35b9e2eb4e // indirect
 	google.golang.org/genproto v0.0.0-20211223182754-3ac035c7e7cb // indirect
-	google.golang.org/grpc v1.41.0
+	google.golang.org/grpc v1.42.0
+	k8s.io/client-go v0.23.0
+	sigs.k8s.io/controller-runtime v0.11.1
 )
+
+replace github.com/tinkerbell/tink => github.com/micahhausler/tink v0.0.0-20220405214824-8387611ceaea
```
displague, line 36: What PR's do we depend on to remove this `replace`?

hardware/kubernetes_client.go
```diff
@@ -0,0 +1,255 @@
...
+	Hostname        string                                               `json:",omitempty"`
+	Plan            string                                               `json:",omitempty"`
+	Factility       string                                               `json:",omitempty"`
+	Tags            []string                                             `json:",omitempty"`
+	SSHKeys         []string                                             `json:"ssh_keys,omitempty"`
+	OperatingSystem ExportableK8sHardwareMetadataInstanceOperatingSystem `json:"operating_system,omitempty"`
+	Network         ExportableK8sHardwareMetadataInstanceNetwork         `json:",omitempty"`
+}
+
+type ExportableK8sHardwareMetadataInstanceOperatingSystem struct {
```
displague, line 230: Seems we could be using the existing types for most of this to prevent introducing diff and drift:

https://github.com/tinkerbell/tink/blob/main/pkg/apis/core/v1alpha1/hardware_types.go#L156-L162

hardware/kubernetes_client_test.go
```diff
@@ -0,0 +1,127 @@
...
+	opts := args.Get(2).([]crclient.ListOption)
+	require.Len(t, opts, 1)
+
+	matchingFields, ok := opts[0].(crclient.MatchingFields)
+	require.True(t, ok)
+
+	require.Contains(t, matchingFields, tink.HardwareIPAddrIndex)
+	assert.Equal(t, ip, matchingFields[tink.HardwareIPAddrIndex])
+
+	// todo(chrisdoherty4) Validate the returned hardware Export() has correctly serialized data.
```
displague, line 64: Would a short string comparison should work for this test?

hardware/kubernetes_client.go
```diff
@@ -0,0 +1,255 @@
...
+	}
+
+	return KubernetesClientConfig{
+		Config:    config,
+		Namespace: namespace,
+	}, nil
+}
+
+// ExportableK8sHardwareHardwareFromK8sObject creates an ExportableK8sHardware from tinkHardware.
+func ExportableK8sHardwareHardwareFromK8sObject(tinkHardware *tinkv1alpha1.Hardware) *ExportableK8sHardware {
```
displague, line 172: This name (`ExportableK8sHardwareHardwareFromK8sObject`) suggests to me that it could be defined, instead, as:

```go
type K8sHardware struct {}
func (*K8sHardware) FromTinkHardware(*tinkv1alpha1.Hardware)
func (*K8sHardware) Export() ([]byte, error)
```

I kind of wish we could avoid the `Export` name and go with `MarshalJSON` to fulfill `Marshaler`. I didn't look around to see how many places would be affected and whether this would be a breaking change, but I do see that at least one `Export` invocation preexists.

displague (commented):
I took an initial look to get a sense of what's in flux and left a few questions.

.gitattributes
chrisdoherty4: Do you mean `go.mod`? `go.sum` is generated and just a bunch of checksums based on the `go.mod` dependency tree.

chrisdoherty4: If we don't want this I can remove it. I've done this on past projects to remove misleading line counts.

.golangci.yml
chrisdoherty4: The root project does, but `golangci-lint` has it on by default. See https://golangci-lint.run/usage/linters/#errorlint. I had Github action errors as a result.

go.mod
chrisdoherty4: https://github.com/tinkerbell/tink/pull/573

hardware/kubernetes_client.go
chrisdoherty4: The existing types have a different structure so while they seem similar, they don't satisfy the Hegel metadata expectations from the `Export()` call. 

Moreover, I think its beneficial to have a clear distinction between Hegel metadata structure and back-end storage format as it allows us to change the structure of the back-end data without impacting hegel given there's an explicit translation.

hardware/kubernetes_client.go
chrisdoherty4: Can use `K8sHardware`. I'm not sure that method should be on the struct though. It happens to use the fields but its not a behavior the struct should provide.

chrisdoherty4: Renamed. Re-open if you'd like to talk methods more.

go.mod
chrisdoherty4: I can't add labels but if you could add a do-not-merge I'll track the other and ping when its merged.

.golangci.yml
mmlb: This file is managed centrally via github.com/tinkerbell/lint-install in an effort to have all the go code passing the same linter settings and thus hopefully look/feel similar. If we're to disable this rule we should do that there. I'm not sure I agree about disabling this btw, if you do send this change over there please explain further how this ends up displaying Hyrums Law.

go.mod
mmlb: I've added the label.

.golangci.yml
chrisdoherty4: Errors are part of the package API. That means if you expose a Kubernetes specific error the Kubernetes specific error is part of the packages API whether you want it or not. That should really be a conscious decision by the package author, not mandated by a linter.

Hyrum's law comes into play because consumers become dependent on that error type and that was never intended making things harder to change in the future.

It's less relevant for internal packages but I typically advocate for the rationale above anyway.

chrisdoherty4: Fwiw errorlint even calls this out as a trade off.

hardware/kubernetes_client_test.go
chrisdoherty4: I suspect so. I'll put something together.

.gitattributes
displague: I see. `k8s.io/client-go` and `sigs.k8s.io/controller-runtime` were added while a few other dependencies were removed. Quite a large number of indirect dependencies came from this, but I think these shouldn't be problems because Tink has roughly the same dependencies (differing on version).

chrisdoherty4: Yes. I was alarmed myself at the transitive dependencies introduced because it literally doubled but then realized it was mostly due to Kube pkgs.

hardware/kubernetes_client_test.go
chrisdoherty4: Added

hardware/kubernetes.go
chrisdoherty4: Resolving as I don't intend to address. Reopen if needed.

hardware/kubernetes_test.go
chrisdoherty4: Won't address on this PR. Reopen if needed.

.golangci.yml
chrisdoherty4: Removed and raised in the `lint-install` repo.

hardware/kubernetes_client.go
chrisdoherty4: I think this is OK. Please unresolve if needed.

Makefile
```diff
@@ -1,16 +1,30 @@
 binary := cmd/hegel
-all: ${binary}
-.PHONY: ${binary} gen test
-${binary}: grpc/protos/hegel
-${binary}:
-	CGO_ENABLED=0 GOOS=$$GOOS go build -ldflags="-X main.GitRev=$(shell git rev-parse --short HEAD)" -o $@ ./$@
+
+all: build
```
micahhausler, line 3: Needs a `.PHONY`

Makefile
```diff
@@ -1,16 +1,30 @@
...
+	docker build $(IMAGE_ARGS) -f ./cmd/hegel/Dockerfile .
+
+.PHONY: gen
 gen: grpc/protos/hegel/hegel.pb.go
+
 grpc/protos/hegel/hegel.pb.go: grpc/protos/hegel/hegel.proto
 	protoc --go_out=plugins=grpc:./ grpc/protos/hegel/hegel.proto
 	goimports -w $@
+
+unit-test:
```
micahhausler, line 20: Needs a `.PHONY`

hardware/kubernetes_client.go
```diff
@@ -0,0 +1,250 @@
...
+				AddressFamily: ip.Family,
+				Public:        ip.Public,
+			},
+		)
+	}
+
+	return hw
+}
+
+// ExportabelK8sHardware satisfies Hardware such that hardware metadata can be exported and queried.
```
micahhausler, line 203: Nit: comment doesn't match struct name

micahhausler (approved):
LGTM modulo minor nits, and pending merge of tinkerbell/tink#573

go.mod
```diff
@@ -29,10 +29,12 @@ require (
...
```
micahhausler, line 40: @chrisdoherty4 tinkerbell/tink#573 is merged!

go.mod
```diff
@@ -24,36 +23,56 @@ require (
 	github.com/tinkerbell/tink v0.0.0-20201207184750-9c52cca562bc
 	go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc v0.26.1
 	go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp v0.28.0
-	google.golang.org/grpc v1.41.0
+	google.golang.org/grpc v1.42.0
+)
+
+require (
```
micahhausler, line 29: Why is the k8s stuff in a separate require block? Is that autogenerated?

chrisdoherty4: I honestly don't know. I've never seen `go` do this before. The indirect deps appearing separate was an addition in 1.17 I believe but this is peculiar.

chrisdoherty4: It might have been merge conflict stuff actually.

hardware/kubernetes_client.go
```diff
@@ -0,0 +1,250 @@
...
+	// KubernetesClient
+	manager, err := tink.NewManager(config.Config, opts)
+	if err != nil {
+		return nil, err
+	}
+
+	managerCtx, cancel := context.WithCancel(context.Background())
+	go func() {
+		if err := manager.Start(managerCtx); err != nil {
+			panic(err)
```
jacobweinstock, line 55: If this err doesnt manifest until some set of circumstances that are only satisfied when an incoming request is handled, this might be difficult to troubleshoot. Having this called in func main somehow, might be something to consider.

hardware/kubernetes_client.go
```diff
@@ -0,0 +1,250 @@
...
+	// Use a manager from the tink project so we can take advantage of the indexes and caching it configures.
+	// Once started, we don't really need any of the manager capabilities hence we don't store it in the
+	// KubernetesClient
+	manager, err := tink.NewManager(config.Config, opts)
+	if err != nil {
+		return nil, err
+	}
+
+	managerCtx, cancel := context.WithCancel(context.Background())
+	go func() {
```
jacobweinstock, line 53: hmmm...removing this goroutine from this function might be something to consider. The name suggests that it is a constructor and having a constructor start a goroutine seems a bit unusual, could be just me though.

refs: [1](https://github.com/golang/go/wiki/CodeReviewComments#synchronous-functions), [2](https://github.com/golang/go/wiki/CodeReviewComments#goroutine-lifetimes)

hardware/kubernetes_client.go
```diff
@@ -0,0 +1,250 @@
...
+	}
+
+	return KubernetesClientConfig{
+		Config:    config,
+		Namespace: namespace,
+	}, nil
+}
+
+// NewK8sHardwareFromK8sObject creates an K8sHardware from tinkHardware.
+func NewK8sHardwareFromK8sObject(tinkHardware *tinkv1alpha1.Hardware) *K8sHardware {
```
jacobweinstock, line 167: with the input to this func being `*tinkv1alpha1.Hardware` does it need the `FromK8sObject` in its name?

[Go and short names](https://github.com/golang/go/wiki/CodeReviewComments#variable-names)

hardware/kubernetes_client.go
chrisdoherty4: Now that I've plumbed in `oklog/run` I can do this. When it was originally possible this was nasty to do :(

hardware/kubernetes_client.go
chrisdoherty4: Adjusted.

hardware/kubernetes_client.go
chrisdoherty4: Generally I agree. The way the app is designed though makes this difficult to do. The `NewClient()` call is a factory function returning an interface so anything that needs to be done for a given implementation _must_ be done during that call. Without re-designing this factory I don't think there's much I can do.

hardware/kubernetes_client.go
chrisdoherty4: Scratch that. See my [other comment](https://github.com/tinkerbell/hegel/pull/82#discussion_r867362342).

Resolving this in favor of that discussion.

PR #84, opened 2022-04-20
Cpd/refactor cli args

chrisdoherty4:
Builds on #82.

Refactor CLI and environment variables such that configuration is an isolated concern.

This is a draft PR, more detail will be added later.


Makefile
```diff
@@ -1,16 +1,31 @@
...
+	docker build $(IMAGE_ARGS) -f ./cmd/hegel/Dockerfile .
+
+.PHONY: gen
 gen: grpc/protos/hegel/hegel.pb.go
+
 grpc/protos/hegel/hegel.pb.go: grpc/protos/hegel/hegel.proto
 	protoc --go_out=plugins=grpc:./ grpc/protos/hegel/hegel.proto
 	goimports -w $@
+
+unit-test:
```
micahhausler, line 21: does this need a `.PHONY`?

cmd/hegel/Dockerfile
```diff
@@ -1,7 +1,17 @@
-FROM golang:1.17.5-alpine
-COPY . /usr/myapp
-WORKDIR /usr/myapp
-RUN mkdir -p bin && go build -o ./bin ./...
+FROM golang:1.17.5-alpine AS builder
```
micahhausler, line 1: go.mod is set to 1.13, that should get a bump, maybe a separate small PR though

cmd/hegel/main.go
```diff
@@ -24,7 +30,183 @@ var (
...
+type RootCommand struct {
+	*cobra.Command
+	RootCommandOptions
+}
+
+func NewRootCommand() (*RootCommand, error) {
+	rootCmd := &RootCommand{
+		Command: &cobra.Command{
+			Use: os.Args[0],
+			Long: `
```
micahhausler, line 60: Why define the help manually?

cmd/hegel/main.go
```diff
@@ -24,7 +30,183 @@ var (
...
+  --factility              HEGEL_FACILITY
+  --http-port              HEGEL_HTTP_PORT
+  --http-custom-endpoints  HEGEL_HTTP_CUSTOM_ENDPOINTS
+
+For backwards compatibility a set of deprecated CLI and environment variables are still supported. Behavior for
+specifying both deprecated and current forms is undefined.
+
+Deprecated CLI Flags
+  Deprecated   Current
+  --http_port  --http-port
```
micahhausler, line 76: Why not [mark the flags as deprecated](https://pkg.go.dev/github.com/spf13/pflag#FlagSet.MarkDeprecated)?

cmd/hegel/main.go
```diff
@@ -24,7 +30,183 @@ var (
...
+		fmt.Fprintf(os.Stderr, "error: %v", err)
+		os.Exit(1)
+	}
+}
+
 func main() {
+	cmd, err := NewRootCommand()
+	handleMainError(err)
+	handleMainError(cmd.Execute())
+	os.Exit(0)
```
micahhausler, line 208: Why exit here?

datamodel/datamodel.go
```diff
@@ -0,0 +1,9 @@
+/*
+Package datamodel defines constants for the various back-end storage mechanisms supported by hegel.
+*/
+package datamodel
+
+const (
+	TinkServer = "1"
+	Kubernetes = "kubernetes"
+)
```
micahhausler, line 9: Is it worth a type alias?
```suggestion

type DataModel string

const (
	DataModelTinkServer = DataModel("1")
	DataModelKubernetes = DataModel("kubernetes")
)
```

hardware/kubernetes_client.go
```diff
@@ -0,0 +1,255 @@
...
+	}
+
+	return KubernetesClientConfig{
+		Config:    config,
+		Namespace: namespace,
+	}, nil
+}
+
+// NewK8sHardwareFromK8sObject creates an K8sHardware from tinkHardware.
+func NewK8sHardwareFromK8sObject(tinkHardware *tinkv1alpha1.Hardware) *K8sHardware {
```
micahhausler, line 172: Tink's `pkg/convert` has conversions for templates and workflows. I think this is fine here, as we probably want to refactor the `Client` interface here to have some more generic hardware representation rather than Tink's eventually

hardware/kubernetes_client.go
```diff
@@ -0,0 +1,255 @@
...
+	if len(hw.Items) == 0 {
+		return nil, fmt.Errorf("no hardware with ip '%v'", ip)
+	}
+
+	if len(hw.Items) > 1 {
+		names := make([]string, len(hw.Items))
+		for i, item := range hw.Items {
+			names[i] = item.Name
+		}
+		return nil, fmt.Errorf("multiple hardware with ip '%v': [%v]", ip, strings.Join(names, ", "))
```
micahhausler, line 125: This seems like an information leak, maybe don't return the enumerated list? Log it for an admin that cares, but the client doesn't need to know

micahhausler (commented):
This is great, a few questions and minor changes

Makefile
chrisdoherty4: Yes

cmd/hegel/Dockerfile
chrisdoherty4: Agree

cmd/hegel/main.go
chrisdoherty4: To clarify, this doesn't define the complete help manually. It defines a long description that gets used with `-h`. The flags portion is still auto gen.

The purpose is to clarify deprecated vs current arguments and how things map. I felt it was appropriate to include the documentation until we remove things but I'm on the fence on whether its necessary.

hardware/kubernetes_client.go
chrisdoherty4: This is from the other PR, will update there.

datamodel/datamodel.go
chrisdoherty4: Yes

cmd/hegel/main.go
chrisdoherty4: Ignore it, its temporary. This was just for testing purposes to make sure the CLI worked correctly with env vars etc.

I'll probably strip this out and rework the sources on this PR, then introduce the actual changes on a separate PR to keep things isolated and easier to review.

hardware/kubernetes_client.go
chrisdoherty4: I noticed Boots is using similar types. I think we would benefit from some sort of back-end hardware module thats used among all services as we seem to have some dupe effort.

datamodel/datamodel.go
chrisdoherty4: Will add to the other PR

hardware/client.go
```diff
@@ -71,38 +73,47 @@ type watcherTinkerbell struct {
...
 }
 
 // NewClient returns a new hardware Client, configured appropriately according to the mode (Cacher or Tink) Hegel is running in.
 func NewClient() (Client, error) {
-	var hg Client
-
 	switch dataModelVersion {
-	case "1":
+	case datamodel.Kubernetes:
+		kubeAPI := os.Getenv("HEGEL_KUBE_API")
```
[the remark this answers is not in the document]
chrisdoherty4: Yes. Will actually start using the CLI configuration etc in a separate PR I think. Else there's lots going on.

cmd/hegel/main.go
chrisdoherty4: Resolving. Unresolve if further discussion needed.

datamodel/datamodel.go
chrisdoherty4: Reworked PR stack so now done here.

thebsdbox (approved):

PR #86, opened 2022-04-22
Wire cobra into Hegel

chrisdoherty4:
## Overview

Builds on #84.

Wire Cobra into the business logic and strip the existing environment reads and flag parsing from the code base centralizing configuration. In wiring the CLI with the business logic I took the liberty of removing the `runGoroutine` function in favor of a small, self-contained library called `oklog/run`. Its only purpose is to manage sets of goroutines that should run collectively like those created during a main process (see root_cmd.go). There are no changes to existing behavior.

Patched a bug where the http server never shuts down after the container receives an interrupt/termination signal.

Additionally, I've rejigged the Dockerfile to take advantage of image layer caching having experienced difficulty when operating on corporate networks.

## How has this been tested?

Unit tests pass and the service run up stand-alone.

### DIFF
--- a/build/build.go
+++ b/build/build.go
@@ -1,0 +1,11 @@
+// Package build contains build specific information.
+package build
+
+// gitRevision is injected at build time.
+var gitRevision string
+
+// GetGitRevision retrieves the revision injected into the build at build time.
+// Deprecated. Removed when moving to 1.18 in favor of runtime/debug.ReadBuildInfo().
+func GetGitRevision() string {
+	return gitRevision
+}
--- a/cmd/hegel/root_cmd.go
+++ b/cmd/hegel/root_cmd.go
@@ -1,15 +1,28 @@
 package main
 
 import (
-	"errors"
 	"fmt"
 	"os"
+	"os/signal"
 	"strings"
+	"syscall"
+	"time"
 
+	_ "net/http/pprof" //nolint:gosec // G108: Profiling endpoint is automatically exposed on /debug/pprof
+
+	"github.com/equinix-labs/otel-init-go/otelinit"
+	"github.com/oklog/run"
+	"github.com/packethost/pkg/log"
+	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
 	"github.com/spf13/viper"
+	"github.com/tinkerbell/hegel/build"
 	"github.com/tinkerbell/hegel/datamodel"
+	grpcserver "github.com/tinkerbell/hegel/grpc-server"
+	"github.com/tinkerbell/hegel/hardware"
+	httpserver "github.com/tinkerbell/hegel/http-server"
+	"github.com/tinkerbell/hegel/metrics"
 )
 
 const longHelp = `
@@ -29,17 +42,18 @@ specifying both deprecated and current forms is undefined.
 Deprecated CLI flags
   Deprecated   Current
   --http_port  --http-port
-  --use_tls    --grpc-use-tls
   --tls_cert   --grpc-tls-cert
   --tls_key    --grpc-tls-key
+  --use_tls    --grpc-use-tls
 
 Deprecated environment variables
   Deprecated          Current
+  CUSTOM_ENDPOINTS    HEGEL_HTTP_CUSTOM_ENDPOINTS
   DATA_MODEL_VERSION  HEGEL_DATA_MODEL
+  GRPC_PORT           HEGEL_GRPC_PORT
   HEGEL_TLS_CERT      HEGEL_GRPC_TLS_CERT
   HEGEL_TLS_KEY       HEGEL_GRPC_TLS_KEY
-  USE_TLS             HEGEL_GRPC_USE_TLS
-  CUSTOM_ENDPOINTS    HEGEL_HTTP_CUSTOM_ENDPOINTS
+  HEGEL_USE_TLS       HEGEL_GRPC_USE_TLS
   TRUSTED_PROXIES     HEGEL_TRUSTED_PROXIES
 `
 
@@ -48,16 +62,17 @@ const EnvNamePrefix = "HEGEL"
 
 // RootCommandOptions encompasses all the configurability of the RootCommand.
 type RootCommandOptions struct {
-	Facility       string `mapstructure:"facility"`
 	DataModel      string `mapstructure:"data-model"`
+	Facility       string `mapstructure:"facility"`
 	TrustedProxies string `mapstructure:"trusted-proxies"`
 
-	HTTPPort            int    `mapstructure:"http-port"`
 	HTTPCustomEndpoints string `mapstructure:"http-custom-endpoints"`
+	HTTPPort            int    `mapstructure:"http-port"`
 
-	GrpcTLSCertPath string `mapstructure:"grpc-tls-cert"`
-	GrpcTLSKeyPath  string `mapstructure:"grpc-tls-key"`
-	GrpcUseTLS      bool   `mapstructure:"grpc-use-tls"`
+	GRPCPort        int    `mapstructure:"grpc-port"`
+	GRPCTLSCertPath string `mapstructure:"grpc-tls-cert"`
+	GRPCTLSKeyPath  string `mapstructure:"grpc-tls-key"`
+	GRPCUseTLS      bool   `mapstructure:"grpc-use-tls"`
 
 	KubeAPI    string `mapstructure:"kubernetes-api"`
 	Kubeconfig string `mapstructure:"kubeconfig"`
@@ -70,7 +85,7 @@ func (o RootCommandOptions) GetDataModel() datamodel.DataModel {
 // RootCommand is the root command that represents the entrypoint to Hegel.
 type RootCommand struct {
 	*cobra.Command
-	Vpr  *viper.Viper
+	vpr  *viper.Viper
 	Opts RootCommandOptions
 }
 
@@ -91,8 +106,8 @@ func NewRootCommand() (*RootCommand, error) {
 	rootCmd.Flags().SortFlags = false // Print flag help in the order they're specified.
 
 	// Ensure keys with `-` use `_` for env keys else Viper won't match them.
-	rootCmd.Vpr = viper.NewWithOptions(viper.EnvKeyReplacer(strings.NewReplacer("-", "_")))
-	rootCmd.Vpr.SetEnvPrefix(EnvNamePrefix)
+	rootCmd.vpr = viper.NewWithOptions(viper.EnvKeyReplacer(strings.NewReplacer("-", "_")))
+	rootCmd.vpr.SetEnvPrefix(EnvNamePrefix)
 
 	if err := rootCmd.configureFlags(); err != nil {
 		return nil, err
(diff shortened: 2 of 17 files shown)


build/build.go
```diff
@@ -0,0 +1,12 @@
+/*
+Package build contains build specific information.
+*/
```
mmlb, line 3: can you do `//` instead, `/* */` aren't usually used in go and definitely not in this code base.

build/build.go
```diff
@@ -0,0 +1,12 @@
...
+package build
+
+// gitRevision is injected at build time.
+var gitRevision string
+
+// GetGitRevision retrieves the revision injected into the build at build time.
+func GetGitRevision() string {
```
mmlb, line 10: Can you add a todo comment to use `runtime/debug.ReadBuildInfo` when we no longer support go1.17 please?

cmd/hegel/root_cmd.go
```diff
@@ -105,52 +120,110 @@ func NewRootCommand() (*RootCommand, error) {
...
+	metrics.State.Set(metrics.Initializing)
+
+	hardwareClient, err := hardware.NewClient(c.Opts.Facility, c.Opts.GetDataModel())
+	if err != nil {
+		return fmt.Errorf("create client: %w", err)
+	}
+
+	grpcServer := grpcserver.NewServer(logger, hardwareClient)
+
+	ctx, cancel := signals.ContextCanceller(cmd.Context(), syscall.SIGINT, syscall.SIGTERM)
```
mmlb, line 155: shouldn't this be using `ctx` returned by `otelinit.InitOpenTelementry`?

cmd/hegel/root_cmd.go
```diff
@@ -105,52 +120,110 @@ func NewRootCommand() (*RootCommand, error) {
...
+	var routines run.Group
+
+	routines.Add(
```
mmlb, line 158: would it make more sense to use `g, ctx := errgroup.WithContext(...)`. TIL about oklog/run and gotta say not a fan of the api taking 2 funcs, this seems more go like to me:

```go
ctx, cancel := signals.ContextCanceller(cmd.Context(), syscall.SIGINT, syscall.SIGTERM)
defer cancel()

g, ctx := errgroup.WithContext(ctx)
g.Go(func() error { return httpserver.Serve(...) })
g.Go(func() error { return grpcserver.Serve(...) })
return g.Wait()
```
The whole 2 func thing, one for error handling is mostly the weird part :). Its also nice that we can just defer the cancel asap, although I'm pretty sure we can do that with oklog/run too.

cmd/hegel/root_cmd.go
```diff
@@ -36,9 +49,10 @@ Deprecated CLI flags
 Deprecated environment variables
   Deprecated          Current
   DATA_MODEL_VERSION  HEGEL_DATA_MODEL
+  GRPC_PORT           HEGEL_GRPC_PORT
+  HEGEL_USE_TLS       HEGEL_GRPC_USE_TLS
```
mmlb, line 53: `HEGEL_USE_TLS` should go after `HEGEL_TLS_KEY` to maintain sorted order.

cmd/hegel/root_cmd.go
```diff
@@ -55,9 +69,10 @@ type RootCommandOptions struct {
 	HTTPPort            int    `mapstructure:"http-port"`
 	HTTPCustomEndpoints string `mapstructure:"http-custom-endpoints"`
 
-	GrpcTLSCertPath string `mapstructure:"grpc-tls-cert"`
-	GrpcTLSKeyPath  string `mapstructure:"grpc-tls-key"`
-	GrpcUseTLS      bool   `mapstructure:"grpc-use-tls"`
+	GRPCPort        int    `mapstructure:"grpc-port"`
+	GRPCUseTLS      bool   `mapstructure:"grpc-use-tls"`
```
mmlb, line 73: ditto, re sorting

cmd/hegel/root_cmd.go
```diff
@@ -167,32 +240,32 @@ func (rc *RootCommand) configureLegacyFlags() error {
 
 	for key, envName := range map[string]string{
 		"data-model":            "DATA_MODEL_VERSION",
+		"grpc-use-tls":          "HEGEL_USE_TLS",
```
mmlb, line 243: sort?

hardware/mock/mock.go
```diff
@@ -26,29 +27,26 @@ func (hg HardwareClient) IsHealthy(context.Context) bool {
...
-	switch dataModelVersion {
-	case "1":
-		hw = &hardware.Tinkerbell{}
-
+	switch hg.Model {
+	case datamodel.TinkServer:
+		fmt.Printf("HELLO WORLD\n")
 		if ip != UserIP {
-			return hw, errors.New("rpc error: code = Unknown desc = unexpected end of JSON input")
+			return nil, fmt.Errorf("received non-mock remote ip address: %v", ip)
```
mmlb, line 34: should be `errors.Errof` (this package should have been using "pkg/errors" not "errors" like the rest of the repo).

hardware/mock/mock.go
```diff
@@ -26,29 +27,26 @@ func (hg HardwareClient) IsHealthy(context.Context) bool {
...
+		hw := &hardware.Tinkerbell{}
+		if err := json.Unmarshal([]byte(hg.Data), hw); err != nil {
 			return nil, err
 		}
+
+		return hw, nil
 	default:
 		if ip != UserIP {
-			return &hardware.Cacher{}, errors.New("rpc error: code = Unknown desc = DB is not ready")
+			return &hardware.Cacher{}, fmt.Errorf("received non-mock remote ip address: %v", ip)
```
mmlb, line 45: ditto

signals/signals.go
```diff
@@ -0,0 +1,20 @@
+package signals
```
mmlb, line 1: why aren't just using `signal.NotifyContext`?

mmlb (commented):
Lots of good stuff here! I've got a few bits of feedback and some questions.

build/build.go
chrisdoherty4: I think I got carried away with `/*` for cmds where you typically want formatted docs for examples etc. Thanks.

signals/signals.go
chrisdoherty4: I thought that came out in 1.18 🤦🏻 thanks.

cmd/hegel/root_cmd.go
chrisdoherty4: This isn't sorted, hence custom endpoints is at further down. The reason for this change was for humans given we read top to bottom in English and you likely want to know if TLS is enabled before worrying about the configuration of TLS. That theme is in the options so appears here to. 

I'm good with either approach, let me know.

cmd/hegel/root_cmd.go
chrisdoherty4: `errgroup` is a good alternative with the tradeoff that it expects everything to work with contexts which isn't the case. Many libraries expose methods for performing shutdown. A good example of that is the `net/http.Server` construct that doesn't actually use contexts for shutdown, it exposes a `Shutdown()` (that accepts a context for timeout type behavior) and `Close()`. 

I don't think 2 funcs is a bad thing. Its an API that delegates the behavior rather than assumes its a context that needs cancelling.

Of course, both can be made to work in the same way as the other so I feel like this is 6 of 1, half a dozen of the other. I'm good to use either.

hardware/mock/mock.go
chrisdoherty4: Do we really feel the need for `pkg/errors`? Its largely redundant since `fmt.Errorf` was introduced unless we feel stack traces are super important (spoiler: I don't). I've always found well written errors with appropriate context suffice.

cmd/hegel/root_cmd.go
chrisdoherty4: I ordered it. Resolved.

cmd/hegel/root_cmd.go
chrisdoherty4: Resolved.

cmd/hegel/root_cmd.go
chrisdoherty4: Resolved.

signals/signals.go
chrisdoherty4: Resolved.

cmd/hegel/root_cmd.go
mmlb: right, but at this point I think its fair to say "context is the way". Hmm actually I'm now seeing that http servers don't have a context for shutting down and that seems annoying. Meh either way then too. Since this is already written then lets keep it. NM.

build/build.go
mmlb: In go1.18 BuildInfo will return vcs for example:

```go
package main

import (
	"fmt"
	"runtime/debug"
)

func main() {
	var gitCommitID string
	var gitCommitTS string
	info, _ := debug.ReadBuildInfo()
	for _, setting := range info.Settings {
		fmt.Printf("setting: name=%s value=%s\n", setting.Key, setting.Value)
		switch setting.Key {
		case "vcs.revision":
			gitCommitID = setting.Value
		case "vcs.time":
			gitCommitTS = setting.Value
		}
	}
	fmt.Println("git commit:", gitCommitID)
	fmt.Println("git commit time:", gitCommitTS)
}
```

```ShellSession
[16:10:42]-[~/go/src/github.com/tinkerbell/boots]─[manny@zennix]> go build -o t; ./t
setting: name=-compiler value=gc
setting: name=CGO_ENABLED value=1
setting: name=CGO_CFLAGS value=
setting: name=CGO_CPPFLAGS value=
setting: name=CGO_CXXFLAGS value=
setting: name=CGO_LDFLAGS value=
setting: name=GOARCH value=amd64
setting: name=GOOS value=linux
setting: name=GOAMD64 value=v1
setting: name=vcs value=git
setting: name=vcs.revision value=08af73b8675e50c27e3422daff5835fab3ddbb90
setting: name=vcs.time value=2022-04-27T20:04:23Z
setting: name=vcs.modified value=true
git commit: 08af73b8675e50c27e3422daff5835fab3ddbb90
git commit time: 2022-04-27T20:04:23Z
```

So when we switch to 1.18+ only we can do away with passing vcs info via compiler/linker flags and just pull it out from debug.

hardware/mock/mock.go
mmlb: If nothing else its important to keep in the style with the rest of the repo, but I do think stacks are nice and use them all the time to hone in while debugging production or even tests.

cmd/hegel/root_cmd.go
chrisdoherty4: We can always change later. I'm not opposed to rallying around errgroup soon.

mmlb: Yeah, agree. I think I do have a branch laying around somewhere for tink I think that used errgroup. I'll compare with `run` whenever I get to it.

build/build.go
chrisdoherty4: Thats nifty. I hadn't seen that in the release notes, thanks.

hardware/mock/mock.go
chrisdoherty4: I took the liberty of replacing all occurrences of stdlib error and `fmt.Errorf`.

build/build.go
chrisdoherty4: Resolved

build/build.go
chrisdoherty4: Resolved.

cmd/hegel/root_cmd.go
chrisdoherty4: Resolved.

hardware/mock/mock.go
chrisdoherty4: Resolved.

hardware/mock/mock.go
chrisdoherty4: Comically, `pkg/errors` doesn't support `%w` necessitating https://github.com/tinkerbell/lint-install/pull/41 😆

mmlb (approved):

mmlb:
d'oh needs a rebase on main

mmlb:
> d'oh needs a rebase on main

nm looks like mergify does the deed

hardware/mock/mock.go
mmlb: Youch, good catch!

PR #95, opened 2022-05-02
Remove pkg level state

chrisdoherty4:
_The branch got messed up during a rebase and Github automatically closed the previous PR (#87). All comments from the previous PR have been addressed._

The primary purpose of this PR is to strip package level state from the service. Global state contributes to application complexity that is considered the cause for many bugs. Package state is 1 notch behind global state.

Package state has been replaced with explicit dependency declarations in functions and, consequently, injected dependencies. This lays a foundation for further refactoring to bring about cleaner code that's easier to work with. 

With the state removal some additional fixes and restructuring was done; specifics:

- all http handlers are now high-order functions returning an `http.Handler` that leverages the injected dependencies.
- metrics gathered by polling the health function of clients has been isolated in the `metrics` package and is launched as a distinct goroutine. 
- a series of bugs in the http health check handler have been fixed. Namely continuing on to write payloads after `http.StatusInternalServerError` codes are written. 
- removal of redundant `Content-Type` headers as the setter was called _after_ `WriteHeader()` meaning it has no effect.
- changed the version end-point to return a _known_ JSON structure in contrast to the previous implementation that, presumably, relied on build time information being injected; the structure used is modeled after the health check endpoint and uses `git_rev` as the JSON key.
- changed `http-server` package to `http` and `grpc-server` package to `grpc`; these changes have been isolated on their own commit for ease of reviewing.

Metric package state is untouched given it'll likely require a more thought out refactor than simply removing the state and injecting a dependency.

Similarly, I've purposely avoided writing additional tests as I'd like to do a hoslitic rework of unit and e2e testing.

The only behavioral change is that the `/_packet/version` endpoint payload is now structured and _may_ be different to existing expectations.

### DIFF
--- a/http/server_test.go
+++ b/http/server_test.go
@@ -1,4 +1,4 @@
-package httpserver
+package http
 
 import (
 	"context"
@@ -16,7 +16,7 @@ import (
 	"github.com/packethost/pkg/log"
 	"github.com/stretchr/testify/require"
 	"github.com/tinkerbell/hegel/datamodel"
-	grpcserver "github.com/tinkerbell/hegel/grpc-server"
+	"github.com/tinkerbell/hegel/grpc"
 	"github.com/tinkerbell/hegel/hardware"
 	"github.com/tinkerbell/hegel/hardware/mock"
 	_ "github.com/tinkerbell/hegel/metrics" // Initialize metrics.
@@ -892,7 +892,7 @@ func TestServe(t *testing.T) {
 	customEndpoints := `{"/metadata":".metadata.instance"}`
 
 	go func() {
-		if err := Serve(context.Background(), logger, mock.HardwareClient{}, &grpcserver.Server{}, mport, time.Now(), "", customEndpoints, ""); err != nil {
+		if err := Serve(context.Background(), logger, mock.HardwareClient{}, &grpc.Server{}, mport, time.Now(), "", customEndpoints, ""); err != nil {
 			t.Errorf("Serve() error = %v", err)
 		}
 	}()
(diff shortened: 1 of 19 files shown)


jacobweinstock (dismissed):
lgtm

http/server_test.go
```diff
@@ -15,36 +14,30 @@ import (
...
-	hc := mock.HardwareClient{}
-	hegelServer = grpcserver.NewServer(logger, hc)
-
-	os.Exit(m.Run())
-}
-
 // TestTrustedProxies tests if the actual remote user IP is extracted correctly from the X-FORWARDED-FOR header according to the list of trusted proxies provided.
 func TestTrustedProxies(t *testing.T) {
+	logger, err := log.Init(t.Name())
+	require.NoError(t, err)
```
micahhausler, line 29: This and elsewhere can be replaced with

```suggestion
	logger := log.Test(t, t.Name())
```

micahhausler (dismissed):
LGTM. Just one minor change on test log initialization

displague:
> The only behavioral change is that the /_packet/version endpoint payload is now structured and may be different to existing expectations.

@mmlb is this potentially a problem?
I wonder if a new endpoint would be preferable for this change.

chrisdoherty4:
> > The only behavioral change is that the /_packet/version endpoint payload is now structured and may be different to existing expectations.
> 
> @mmlb is this potentially a problem?
> I wonder if a new endpoint would be preferable for this change. 

As part of 0.7 id like to see these deprecated anyway, removed in 0.8. The endpoint retrieved its data from a variable that didn't have any definition other than indicating it was JSON. The variable was unset suggesting it was set during build time so I felt comfortable removing that instead returning a defined payload.

The fact this is v0 and that this data had no structure to it, it felt appropriate to strip it out.

jacobweinstock (approved):

PR #101, opened 2022-05-09
Fix exposure of locks externally from the grpc server struct

### LINKED ISSUE #93: Handling of subscription query could segfault
See https://github.com/tinkerbell/hegel/pull/87#discussion_r860835171

Retrieving subscriptions doesn't work for a particular case and likely segfaults. Given there's been no complaints it suggests its not being used. Investigate, verify and remove if broken.

chrisdoherty4:
Fixes #93 that we originally thought could segfault. The `json` package actually returns an error about an unsupported type. 

Nonetheless, the issue with lock exposure and management external to an owning struct is dangerous and has been fixed. The endpoint that passed an incorrect type has removed given no-body could've been using it usefully.

### DIFF
--- a/grpc/server.go
+++ b/grpc/server.go
@@ -31,8 +31,8 @@ type Server struct {
 	log            log.Logger
 	hardwareClient hardware.Client
 
-	subLock       *sync.RWMutex
-	subscriptions map[string]*Subscription
+	subscriptionMu *sync.RWMutex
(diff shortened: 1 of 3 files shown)


micahhausler (dismissed):
Changes LGTM

jacobweinstock (dismissed):

jacobweinstock (approved):

PR #105, opened 2022-05-12
Default to in-cluster config and remove kube config requirements

chrisdoherty4:
The Kubernetes configuration currently required via the CLI isn't strictly required. If we aren't in-cluster and nothing was supplied it will naturally fail on loading. This change removes the checks entirely and lets the client creation process error out naturally.


jacobweinstock (approved):

micahhausler (approved):
LGTM

PR #106, opened 2022-05-13
Add kubernetes namespace override

chrisdoherty4:
When launching Hegel in Kubernetes it can default, if no Kubeconfig is specified, to using the ServiceAccount configured with the Pod. The ServiceAccount is the same as the Pod deployment namespace and consequently Hegel defaults to reading from its own namespace. This isn't always desirable as users may want to place the `Hardware` custom resource in a different namespace.

By defaulting to in-cluster config and requiring an operator to configure RBAC and explicitly set the `--kube-namespace` override we have a 'secure and predictable' out of the box deployment model.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ certs/
 out/
 *.coverage
 /hegel
+.vscode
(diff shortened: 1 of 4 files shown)


micahhausler (approved):
LGTM

jacobweinstock (approved):

PR #110, opened 2022-06-02
go.mod: Update tink client

mmlb:
## Description

Bumps version of github.com/tinkerbell/tink to get an updated tinkerbell/tink/client package.

## Why is this needed

This is to be able to connect to tink that dropped the /cert endpoint and relaxed TLS requirement.

## How Has This Been Tested?

`go build` and `go test`ed.

Can connected to current versions of tink-server.


mmlb:
nevermind, I had an outdated `main` and didn't notice this has already been updated.

PR #122, opened 2022-09-01
Handle gRPC shutdown from context cancel

jacobweinstock:
## Description

This fixes an issue where a single ctrl-c to hegel doesn't shutdown the gRPC server and requires another ctrl-c in order for hegel to exit at all. This causes deployments into envs like Kubernetes to have to be force-ably killed in order to shutdown.

## Why is this needed

Fixes: #

## Checklist:

I have:

- [ ] added unit or e2e tests

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -16,7 +16,7 @@ IMAGE_ARGS ?= -t hegel
 
 .PHONY: image
 image:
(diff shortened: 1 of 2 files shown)


chrisdoherty4 (approved):

chrisdoherty4:
@mergifyio queue

PR #131, opened 2022-10-29
Remove custom endpoints

### LINKED ISSUE #132: Remove custom endpoints config
Custom endpoint configuration requires operators to understand the implementation detail on a per backend basis. Exposing/requiring operators to understand service implementation is a bad pattern to follow and creates transparent transitive dependencies making for a hard to maintain code base.

Removing the custom endpoints until they're actually required and exposing EC2 metadata endpoints/other known APIs is both simpler and easier to maintain.

chrisdoherty4:
## Overview

Custom endpoints force operators to understand the format exported by the backend so they can write a query to extract the data to serve from the custom endpoint.

Removing the custom endpoint configuration is necessary for further refactoring to remove ambiguous data 'export' from the backend and cleans the code up so operators do not need to understand implementation detail.

In removing custom endpoints, the `registerCustomEndpoints` func was removed also. This in turn rendered many unit tests redundant as they dependend on the func to perform arbitrary configuration for test. The tests themselves made little sense given they configured a mux and didn't really test the application.

The default `/metadata` endpoint has been hard coded with the other mux configuration.

## Next steps

Having removed the custom endpoints, we can refactor the internals as there's no obligation to honor existing 'exports' within the 'backend' code. The intent is to remove ambiguity in the implementation and firm up expectations for each endpoint with respect to where the data comes from (i.e. strip out the ambiguous 'export' from backends in favor of something explicit).

Closes #132


displague:
I took a look at this and it looked fine. The one failing check is around a loss of coverage which makes sense with the tested code removal.

I would like @mmlb and @nshalman to weigh-in because the removal of these custom endpoints would have an effect in Equinix Metal were this code in use (metadata / components for example).

chrisdoherty4:
> how does this affect Tinkerbell /userdata requests in EC2 mode

The `/user-data` endpoint (part of the EC2 standard) is served in the same way as the other EC2 endpoints. That means this change has no impact on it. The `/metadata` endpoint is an Equinix specific thing that I'm looking to maintain in the interim but will likely be dropped in the future too.

See https://github.com/tinkerbell/hegel/blob/main/http/handlers.go#L28

> I would like @mmlb and @nshalman to weigh-in because the removal of these custom endpoints would have an effect in Equinix Metal were this code in use

@nshalman has advised Equinix plan to take a different direction so Hegel's dominant use-case is Tinkerbell. As such, this and many more draft refactors are focused on stripping out anything non-standard/Equinix specific. 

Ref: https://cloud-native.slack.com/archives/C01SRB41GMT/p1658236374550009

@displague

jacobweinstock (approved):

jacobweinstock (approved):

PR #139, opened 2022-10-31
Refactor http handlers and instrument graceful shutdown

chrisdoherty4:
This change set seeks to separate HTTP handler configuration from HTTP server configuration and serving. Separating the handlers from the server that serves the endpoints makes the endpoint testing much easier as it doesn't require real HTTP server's, therefore no port juggling. 

To create a better separation the following changes were made:
- Create a `handler` package with a single API: `New`. The `New` function constructs the endpoints that need exposing based on a parameter. This facilitates the Hegel API that was under development.
- Handler construction is now invoked in the `main` func.
- Middleware configuration has been pulled up to the `main` func.
- `http.Serve` is still called from `main` but no longer launched in a separate goroutine. 
- `http.Serve` now facilitates proper graceful shutdown.

A suite of tests have been deleted that were targetting internal APIs. When evaluating the tests it became evident that they weren't really testing anything useful as we had to construct the application in a consistent way for the tests to be useful. Additional handler tests will be written and placed in `ec2_test.go` or wherever that lands.

Similarly, some tests that were valuable, namely the XFF tests, were re-written and moved to the `XFF` package. The tests are much more thorough in establishing expected XFF middleware behavior.

#### Next steps

#143 begins the process of further splitting apart handlers and removing the ambiguity created from the 'export' concept in the backend implementations. This may result in a further separation of packages as #143 indicates with its `/internal/http/ec2` package.


chrisdoherty4:
@mergifyio queue

PR #143, opened 2022-11-02
Refactor backends and frontends

chrisdoherty4:
### Overview 

This PR introduces a new frontend-backend architecture that facilitates development and exposure of independent frontends and backends. Its primary goal is to bring about clarity in data structures transiting the backend-frontend logical delineation compared with the existing code that has lots of ambiguity and, consequently, is hard to maintain.

### Design

The design lets frontends define the data models they require to serve their endpoints. Their data models are retrieved via a client they define. Backends must satisfy the expected frontend client interface to be compatible with the frontend. This is in contrast to a traditional software architecture where 'domain logic' is generally unaware of ports and adapters.

This design was chosen for its simplicity. When exploring a more traditional approach it became clear that the core domain of Hegel was really around frontends. To make clear exactly what data a frontend needed I defined the data models in the frontend packages. This ensures redundant fields in a 'shared' data model would never exist with the tradeoff of requiring backends to know how to format data for a particular frontend. With this in mind, we can consider backends 'adapters' and frontends as containing the core logic for exposing data in an API compliant way.

<p align="center">

</p>

### Additional work

- Add a help to the Makefile
- Add some helper utilities for testing. They might not be used here but will be used in a separate PR.

### Future work

- Integrate the changes with the root cmd. This work was already done but posterity I'll isolate the integration on a separate PR.
- Writing unit tests.


internal/backend/kubernetes/backend.go
```diff
@@ -0,0 +1,178 @@
+package kubernetes
```
chrisdoherty4, line 1: This file is largely copied from /internal/hardware/kubernetes.go. The only noteworthy difference is the replacement of the `ByIP` method with `GetEC2Instance` given we need to satisfy the `/internal/frontend/ec2.Client` interface.

internal/backend/kubernetes/backend.go
```diff
@@ -0,0 +1,178 @@
...
+
+// IsHealthy returns true until k.Close() is called. It is thread safe.
+func (b *Backend) IsHealthy(context.Context) bool {
+	b.closeM.RLock()
+	defer b.closeM.RUnlock()
+	return b.close != nil
+}
+
+// GetEC2InstanceByIP satisfies ec2.Client.
+func (b *Backend) GetEC2Instance(ctx context.Context, ip string) (ec2.Instance, error) {
```
chrisdoherty4, line 106: This is the additional method that replaces `ByIP`.

go.mod
```diff
@@ -28,6 +28,11 @@ require (
 	sigs.k8s.io/controller-runtime v0.11.1
 )
 
+require (
```
displague, line 31: this is a strange format for go.mod - three require blocks?

.golangci.yml
```diff
@@ -31,6 +31,7 @@ linters-settings:
   errorlint:
     # these are still common in Go: for instance, exit errors.
     asserts: false
+    errorf: false
```
displague, line 34: could we have ignored this on a single line? `//nolint:...` ?
which change prompted this?

chrisdoherty4: The lint rule forces the use of `%w` in error messages. That's a bad thing because you don't always want the original error to be exposed from your package API. For example, if I return a Kubernetes specific error from my Kube backend and a frontend becomes dependent on that, then we switch to a different backend that doesn't throw that same error we have misaligned behavior.

Allowing the use of `%v` reduces the scope of Hyrum's law.

internal/backend/flatfile/backend_test.go
```diff
@@ -0,0 +1,59 @@
...
+	"context"
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
+	. "github.com/tinkerbell/hegel/internal/backend/flatfile"
+	"github.com/tinkerbell/hegel/internal/frontend/ec2"
+)
+
+func TestGetEC2Instance(t *testing.T) {
+	data := struct {
```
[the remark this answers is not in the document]
chrisdoherty4: I think a test case affirming what happens if the IP doesn't exist would be useful. 

I think we probably should isolate `FromYAMLFile` specific tests to a separate root test func as that's not the goal of these tests.

go.mod
chrisdoherty4: I don't know why the Go toolchain does this currently, perhaps its a bug upstream. Either way, I've given up trying to make it consolidate.

I've corrected it here but know its not a manually introduced error.

internal/backend/flatfile/backend_test.go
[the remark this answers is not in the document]
chrisdoherty4: Added a few more test cases.

PR #145, opened 2022-11-03
Improve CI build time

### LINKED ISSUE #135: Use go to cross compile for arm64
The Go toolchain provides great cross compilation support. We need to replace QEMU environment emulation builds that are rather slow with cross compilation builds and just package the artifacts appropriately in an image.

chrisdoherty4:
The CI took in excess of 15m to run per commit. This was largely due to environment emulation used to build binaries and docker images for the supported platforms and serial execution of validation tools. 

This change set does the following:
- Cross compile binaries using the Go toolchain native support.
- Perform multi platform builds using a matrix build that parallelizes the builds.
- Run `go test` as a dedicated test job.
- Cache Go dependencies using the `setup-go` v3 caching feature.
- Updates all actions to their latest versions to ensure we don't begin failing due to various deprecations.

These changes results in an average build time reduction of roughly 600% from 19.5m to 3m (we should really think about using this setup across Tinkerbell).

To provide a sane local developer workflow the `make image` recipe has been updated to always build Linux binaries so it can be packaged into the final image. This means the binaries are no longer built in a container. The `make build` recipe is customizable using familiar `GOOS` and `GOARCH` variables defaulting to whatever your local platform is.

We may consider improving this in the future to limit version skew but given the reliability of the Go toolchain and Go being the only real dependency it seems OK to use the host environment for now. If more developers begin contributing we may need to normalize environments.

#### Minor bits

- Removed the Drone CI related Makefile recipes.
- Changed mergify to only require the `images` job given that job can't run without its pre-req jobs (all other jobs).

Closes #135

### DIFF
--- a/.github/mergify.yml
+++ b/.github/mergify.yml
@@ -3,8 +3,8 @@ queue_rules:
     conditions:
       # Conditions to get out of the queue (= merged)
       - check-success=DCO
+      # The docker-images job only runs if lint, build and test all pass.
       - check-success=images
-      - check-success=validation
 
 pull_request_rules:
   - name: Automatic merge on approval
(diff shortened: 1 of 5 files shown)


jacobweinstock (approved):

jacobweinstock (approved):

chrisdoherty4:
@mergifyio queue

PR #147, opened 2022-11-04
Refactor release workflow

chrisdoherty4:
The [`actions/create-release`](https://github.com/actions/create-release) Github action has been archived. This change set moves to a different action that creates releases in a similar way to the current process.

The action used to create releases is https://github.com/ncipollo/release-action. 

I tested the release workflow with a tag pushed to the Hegel repository and everything works fine. [Successful run can be found under actions](https://github.com/tinkerbell/hegel/actions/runs/3394585868).

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -1,8 +1,8 @@
+name: Release
 on:
   push:
     tags:
       - "v*"
-name: Create release
 
 env:
   REGISTRY: quay.io
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -10,13 +10,13 @@ env:
 
 jobs:
   release:
+    name: Release
     runs-on: ubuntu-latest
     steps:
-      - name: Checkout code
-        uses: actions/checkout@v2
+      - uses: actions/checkout@v3
 
       - name: Login to quay.io
-        uses: docker/login-action@v1
+        uses: docker/login-action@v2
         if: ${{ startsWith(github.ref, 'refs/heads/main') }}
         with:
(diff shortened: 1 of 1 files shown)


jacobweinstock (approved):

mmlb:
hook uses https://github.com/softprops/action-gh-release in https://github.com/tinkerbell/hook/blob/main/.github/workflows/ci.yaml#L144. I don't remember why I used softprops instead of ncipollo's but there was likely a reason. Worth a look to see if the project as a whole can use just 1 releaser.

displague:
Both projects are suggested as alternatives in https://github.com/actions/create-release#github-action---releases-api

Between softprops and ncipollo, the latter appears to be more active, while the former has more forks and stars. 🤷‍♂️ 

Using the same project throughout Tinkerbell would be nice, but this PR doesn't change that existing state nor prevent future movement in that direction.

displague (approved):

displague:
I didn't see an image push in the successful build. Let's make sure the next release actually pushes.

chrisdoherty4:
I never tried out the softprops one, but I compared what ncipollo produces with what we produce today and they were very similar (we no longer produce "new contributors" with this change).

I'm not particularly bothered which solution we use long term, the primary goal of the PR is to get off an archived and clearly unsupported action.

> I didn't see an image push in the fork's successful build. Let's make sure the next release pushes.

It was purposely commented out during those test runs because it relies on a successful build from mainline (as it did previously). Given this hasn't been merged yet, there is no successful build from mainline to release, hence it would [fail in the normal way](https://github.com/tinkerbell/hegel/actions/runs/3394529342/jobs/5643205177).

PR #150, opened 2022-11-07
Remove conditional login on release workflow

chrisdoherty4:
Stop conditionally logging into quay.io on release workflows.

Release workflows are gated on tags so the need to gate the login is superfluous and leads to misleading 400 Bad Request errors from quay.io.


jacobweinstock (approved):

chrisdoherty4:
@mergifyio dequeue

PR #155, opened 2022-11-08
Integrate frontend and backend refactor

chrisdoherty4:
Based on #156 

In https://github.com/tinkerbell/hegel/pull/143 we refactored Hegel to have clearer backends and frontends. This PR integrates that work removing the old code.

The integration has a few intention effects that aren't necessarily obvious from looking at the code.

1. The Tink backend has been dropped. There was little point in converting the Tink backend as we are deprecating it in favor of the Kubernetes backend. Other Tinkerbell services still support this but will be updated in the coming months.
2. The `/versionz` endpoint has been dropped. This endpoint output a build version (vcs revision) only and that information is available on the `/healthz` endpoint also so it doesn't make sense to continue outputting it. A future PR will better document the expectations of the `/versionz` endpoint. While this would ideally be done in a separate PR, it meant converting it to a new handler only to be dropped straight away.
3. The Hegel API no longer functions. The Hegel API wasn't used anywhere and can be converted as desired in a future PR. 

A [separate PR](https://github.com/tinkerbell/hegel/pull/157) has been raised to provide a level of e2e functionality confidence.


internal/frontend/ec2/frontend.go
```diff
@@ -38,7 +38,7 @@ func New(client Client) Frontend {
 // Configure configures router with the supported AWS EC2 instance metadata API endpoints.
 //
 // TODO(chrisdoherty4) Document unimplemented endpoints.
-func (f Frontend) Configure(router *gin.Engine) {
+func (f Frontend) Configure(router gin.IRouter) {
```
jacobweinstock, line 41: nice

jacobweinstock (approved):

PR #157, opened 2022-11-11
Add e2e tests for Hegel root cmd

chrisdoherty4:
Builds on #155.

We have very little automated coverage ensuring the root cmd strings together properly. This change introduces an e2e package that builds and launches the root Hegel command as if a `main()` function would, then runs a few basic tests to ensure it serves data correctly.

The tests are not run as part of the unit test suite and must be invoked with the `-tags=e2e` argument to `go test` as it leverages go build commands to limit how its included.


jacobweinstock (dismissed):
Very nice!

jacobweinstock (approved):

PR #164, opened 2022-11-13
Fix handling of trailing slash in ec2 frontend

### LINKED ISSUE #163: cloud-init integration results in 404s
When cloud-init interrogates the EC2 metadata APIs paths that have nested paths it appends a slash to the query.

E.g. 
```
/2009-04-04/meta-data/
```

Gin won't match these routes as if they didn't have the trailing slash by default resulting in a 404. Consequently, the EC2 API isn't explorable by the cloud-init EC2 datasource implementation.

chrisdoherty4:
cloud-init queries endpoints that nest other endpoints with a trailing slash. Without registering duplicate routes with trailing slashes Hegel will return a 404. This change ensures routes with both a trailing slash and not are registered for the EC2 frontend.

Fixes #163

### DIFF
--- a/internal/ginutil/router.go
+++ b/internal/ginutil/router.go
@@ -1,0 +1,34 @@
+package ginutil
+
+import (
+	"strings"
+
+	"github.com/gin-gonic/gin"
+)
+
+// TrailingSlashRouteHelper wraps a gin.IRouter and ensures every route registered with GET
+// has a corresponding alternate route with or without, dependent on the endpoint being registered,
+// a trailing slash.
+type TrailingSlashRouteHelper struct {
+	gin.IRouter
+}
+
+// GET overrides the internal gin.IRouter GET. It interrogates endpoint for a trailing slash. If
+// no trailing slash is present, it registers the endpoint and a corresponding endpoint with a
+// trailing slash using the same handler. If it does end in a trailing slash, it does the inverse.
+func (r TrailingSlashRouteHelper) GET(endpoint string, handler ...gin.HandlerFunc) gin.IRoutes {
+	// Determine if the alternate endpoint should end with a slash or have it stripped.
+	// This ensures we don't have routes such as
+	// 		/2009-04-04/meta-data/instance-id/
+	// 		/2009-04-04/meta-data/instance-id//
+	var alternateEndpoint string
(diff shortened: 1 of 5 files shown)


internal/ginutil/router.go
```diff
@@ -0,0 +1,34 @@
...
+	gin.IRouter
+}
+
+// GET overrides the internal gin.IRouter GET. It interrogates endpoint for a trailing slash. If
+// no trailing slash is present, it registers the endpoint and a corresponding endpoint with a
+// trailing slash using the same handler. If it does end in a trailing slash, it does the inverse.
+func (r TrailingSlashRouteHelper) GET(endpoint string, handler ...gin.HandlerFunc) gin.IRoutes {
+	// Determine if the alternate endpoint should end with a slash or have it stripped.
+	// This ensures we don't have routes such as
+	// 		/2009-04-04/meta-data/instance-id/
```
displague, line 22: `/2009-04-04/meta-data/instance-id/`

I believe trailing slash is only supported on array and hash tree items. I do not believe a trailing slash should be accepted with scalar values like strings, ints, and bools.

Can you confirm this behavior in EC2?

displague: I was suspecting that on EC2, `GET 169.254.254./2009-04-04/meta-data/instance-id/` would yield a 404.

However, after running against a mocking environment (perhaps not the same, but as close as I can get) this does not appear to be the case.

```
$ nerdctl run -it --rm -p 1338:1338 public.ecr.aws/aws-ec2/amazon-ec2-metadata-mock:v1.11.2
$ curl http://localhost:1338/latest/meta-data/instance-id
i-1234567890abcdef0
$ curl http://localhost:1338/latest/meta-data/instance-id/
i-1234567890abcdef0
```

https://github.com/aws/amazon-ec2-metadata-mock is Apache2 and may have other hints (or types and interfaces).

displague (approved):

internal/ginutil/router.go
chrisdoherty4: Either way, the spec doesn't indicate anything so we're very much left to figure it out. I'm being guided by the things we integrate with - hopefully they never contradict.

PR #167, opened 2022-11-14
Add Kubernetes unit tests

chrisdoherty4:
Add Kubernetes unit tests.


chrisdoherty4:
The coverage drop here stems from the `/internal/backend` package I believe. This package just masks over multiple possible backends providing a convenient way to construct a backend without caring too much about the specifics in the root command.

Maybe we'll add tests one day, maybe we won't 🤷🏻 .

internal/backend/kubernetes/backend.go
```diff
@@ -116,16 +92,71 @@ func (b *Backend) retrieveByIP(ctx context.Context, ip string) (tinkv1.Hardware,
...
+		i.Metadata.Facility = hw.Spec.Metadata.Facility.FacilityCode
+	}
+
+	if hw.Spec.UserData != nil {
+		i.Userdata = *hw.Spec.UserData
+	}
+
+	// TODO(chrisdoherty4) Support public keys. The frontend doesn't handle public keys correctly
+	// as it expects a single string and just outputs that key. Until we can support multiple keys
+	// its not worth adding it to the metadata.
```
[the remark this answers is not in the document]
chrisdoherty4: Its not quite that simple unfortunately. With the IP address case EC2 embeds the idea of exposing only 1. With keys, it embeds the idea of exposing > 1. This manifests in the API as the following:

```
/2009-04-04/meta-data/public-ipv4
/2009-05-05/meta-data/public-keys/0/openssh-key
```

We could hard code the `0/openssh-key` portion of the API but given this has _never_ worked in Hegel I saw it as less of an issue and something that could be addressed separately.

internal/backend/kubernetes/backend.go
```diff
@@ -2,104 +2,80 @@ package kubernetes
...
-}
-
-// WaitForCacheSync waits for the internal Backend cache to synchronize.
-func (b *Backend) WaitForCacheSync(ctx context.Context) bool {
-	return b.waitForCacheSync(ctx)
+	return backend, nil
 }
 
-// IsHealthy returns true until k.Close() is called. It is thread safe.
+// IsHealthy returns true until the context used to create the Backend is cancelled.
```
jacobweinstock, line 61: Is this a step toward something else? basing health of a backend on context feels curious.

internal/backend/kubernetes/backend.go
```diff
@@ -116,16 +92,71 @@ func (b *Backend) retrieveByIP(ctx context.Context, ip string) (tinkv1.Hardware,
...
+		i.Metadata.Tags = hw.Spec.Metadata.Instance.Tags
+
+		if hw.Spec.Metadata.Instance.OperatingSystem != nil {
+			i.Metadata.OperatingSystem.Slug = hw.Spec.Metadata.Instance.OperatingSystem.Slug
+			i.Metadata.OperatingSystem.Distro = hw.Spec.Metadata.Instance.OperatingSystem.Distro
+			i.Metadata.OperatingSystem.Version = hw.Spec.Metadata.Instance.OperatingSystem.Version
+			i.Metadata.OperatingSystem.ImageTag = hw.Spec.Metadata.Instance.OperatingSystem.ImageTag
+		}
+
+		// Iterate over all IPs and set the first one for IPv4 and IPv6 as the values in the
```
jacobweinstock, line 126: how are these ordered? does it depend on the backend? I believe they come back from the backend as a slice? and we dont have any ordering done in kubernetes?

internal/backend/kubernetes/backend.go
```diff
@@ -2,104 +2,80 @@ package kubernetes
...
 }
 
-// IsHealthy returns true until k.Close() is called. It is thread safe.
+// IsHealthy returns true until the context used to create the Backend is cancelled.
 func (b *Backend) IsHealthy(context.Context) bool {
-	b.closeM.RLock()
-	defer b.closeM.RUnlock()
-	return b.close != nil
+	select {
+	case <-b.closer:
```
jacobweinstock, line 64: whats the difference between b.closer and the context passed in? why not just use that one?

internal/backend/kubernetes/config.go
```diff
@@ -1,15 +1,26 @@
...
 	*rest.Config
+
+	// Namespace restricts the scope of the backend such that Hardware objects are retrieved from
+	// this namespace only. Defaults to "default".
 	Namespace string
+
+	// Context is the context used by the Kubernetes client. Defaults to context.Background().
+	// When specified it controls the lifetime of the Kubernetes client by shutting the client
+	// down when it cancelled.
+	Context context.Context
```
jacobweinstock, line 23: curious about your thought process here. context in a struct vs passing the same context around to all func/methods?

internal/backend/kubernetes/backend.go
chrisdoherty4: This is from the context, we're just not holding a reference to the whole context because storing a context felt dirty in itself.

internal/backend/kubernetes/config.go
chrisdoherty4: The struct is just a configuration object. It doesn't do anything but provide data to the `NewBackend`. `NewBackend` doesn't store the whole context, it only stores the `Done()` channel. This is for convenience and to avoid a `Start()` method because, in times gone by, I've found that nearly always results in lots of `NewBackend()` _immediately_ followed by `Start()`. I could go either way though.

If we consider this from the consumers point of view:


```go
func main() {
	ctx, cancel := signals.NotifyContext(context.Background, syscall.SIGUP)
	defer cancel()

	// This
	be := kubernetes.NewBackend(kubernetes.Config{
		// Other config
		Context: ctx,
	})

	// vs this
	be := kubernetes.NewBackend(kubernetes.Config{
		// Other config
	})
	be.Start(ctx)
}

internal/backend/kubernetes/backend.go
chrisdoherty4: Its no different to before. Its not based on a context, its based on what the context represents. The context, in this instance, represents a client with a synchronizing cache that we launch in `NewBackend()`. If that stops, the client is no longer usable, therefore the backend is no longer usable.

I'm split on whether we should trend toward deep or shallow health checks. Which ever way we go, this implementation isn't functionally any different to what we had previously - it just does it in a slightly different way.

internal/backend/kubernetes/backend.go
chrisdoherty4: Kubernetes preserves slice ordering. The contract is whatever is first is chosen.

jacobweinstock (approved):
Thanks for all the explanations!

internal/backend/kubernetes/backend.go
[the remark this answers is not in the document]
displague: IIRC - /2009-04-04/meta-data/public-keys/0/ returned a multiline text document, which cloud-init faithfully lobbed into the authorized_keys, which worked.

Issue #191, opened 2022-11-22
Http server error doesn't terminate Hegel


jacobweinstock:
If Hegel's http server: [`http.Server`](https://github.com/tinkerbell/hegel/blob/e923e5debfec2e81f01c71459d058cddbcbf4c81/internal/http/server.go#L29) errors out, with something like a port in use error, the `Server` function will not return the error or even return from the function. A message will be logged but nothing else and the `Server` function will wait until a signal is received in order to exit from `Server`. This can potential lead to a deployment, in Kubernetes for example, to behave like a successful deployment. This doesn't feel like the desired behavior.

## Expected Behaviour

An error to [`http.Server`](https://github.com/tinkerbell/hegel/blob/e923e5debfec2e81f01c71459d058cddbcbf4c81/internal/http/server.go#L29) will cause an error to be returned from the `Server` function.

## Steps to Reproduce (for bugs)

1.
2.
3.
4.

## Your Environment

* Operating System and version (e.g. Linux, Windows, MacOS):

* How are you running Tinkerbell? Using Vagrant & VirtualBox, Vagrant & Libvirt, on Packet using Terraform, or give details:

* Link to your project or a code example to reproduce issue:

PR #193, opened 2022-11-22
Handle server error

jacobweinstock:
## Description

Return when `server.ListenAndServe()` fails in `http.Serve`. 

## Why is this needed

Fixes: #191 

## Checklist:

I have:

- [ ] added unit or e2e tests

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,6 +4,9 @@ GOARCH ?= $(shell go env GOARCH)
 # Specify the target OS to build the binary for. (Recipes: build)
 GOOS ?= $(shell go env GOOS)
 
+# Specify the GOPROXYs to use in the build of the binary. (Recipes: build, image)
+GOPROXY ?= $(shell go env GOPROXY)
+
 # Specify additional `docker build` arguments. (Recipes: image)
 IMAGE_ARGS ?= -t hegel
 
@@ -19,6 +22,7 @@ build: ## Build the Hegel binary. Use GOOS and GOARCH to set the target OS and a
 	CGO_ENABLED=0 \
 	GOOS=$$GOOS \
 	GOARCH=$$GOARCH \
+	GOPROXY=$$GOPROXY \
 	go build \
 		-o hegel-$(GOOS)-$(GOARCH) \
 		./cmd/hegel
@@ -36,7 +40,7 @@ test-e2e: ## Run E2E tests.
 .PHONY: image
 image: export GOOS=linux
 image: build ## Build a Linux based Hegel image for the the host architecture.
-	docker build --build-arg GOPROXY=$(GOPROXY) $(IMAGE_ARGS) .
+	docker build --build-arg TARGETARCH=$(GOARCH) --build-arg TARGETOS=$(GOOS) $(IMAGE_ARGS) .
 
 # The command to run mockgen.
 MOCKGEN = go run github.com/golang/mock/mockgen@v1.6
--- a/Makefile
+++ b/Makefile
@@ -4,7 +4,7 @@ GOARCH ?= $(shell go env GOARCH)
 # Specify the target OS to build the binary for. (Recipes: build)
 GOOS ?= $(shell go env GOOS)
 
-# Specify the GOPROXYs to use in the build of the binary. (Recipes: build, image)
+# Specify the GOPROXYs to use in the build of the binary. (Recipes: build)
 GOPROXY ?= $(shell go env GOPROXY)
 
 # Specify additional `docker build` arguments. (Recipes: image)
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -36,7 +40,7 @@ test-e2e: ## Run E2E tests.
 .PHONY: image
 image: export GOOS=linux
 image: build ## Build a Linux based Hegel image for the the host architecture.
-	docker build --build-arg GOPROXY=$(GOPROXY) $(IMAGE_ARGS) .
+	docker build --build-arg TARGETARCH=$(GOARCH) --build-arg TARGETOS=$(GOOS) $(IMAGE_ARGS) .
```
chrisdoherty4, line 43: I don't think we need the `TARGETARCH` given this is Make target is meant to build for the host architecture.

internal/http/server.go
```diff
@@ -24,15 +24,21 @@ func Serve(ctx context.Context, logger log.Logger, address string, handler http.
...
 		logger.Info(fmt.Sprintf("Listening on %s", address))
 		if err := server.ListenAndServe(); err != nil && !errors.Is(err, http.ErrServerClosed) {
 			logger.Info(err.Error())
+			errChan <- err
 		}
 	}()
 
 	// Wait until we're told to shutdown.
-	<-ctx.Done()
+	select {
```
chrisdoherty4, line 37: This won't work, unfortunately, because of the pseudorandom case selection behavior of the `select` statement _and_ because we shutdown via `server.Shutdown` on l.50 so, at this point, nothing has told `ListenAndServe` to stop.

To retrieve the error we'll probably want to funnel all errors into `errChan` including from the context timeout. We'd then launch the `Shutdown` in a separate goroutine and retrieve the error from the errChan only.

```go
errChan := make(chan error, 2)

// Launch ListenAndServe

<-ctx.Done()

ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
defer cancel()

go func() {
	errChan <- server.Shutdown(ctx)
}()

// We'll eithe rreceive an error from the ListenAndServe routine or
// a context related error from the timeout.
// For the timeout case we want to forcefully close the server and
// tidy the error but for all other cases we just want to return that error.
//
// Note we don't need a select statement because we _know_ we're using a 
// timeout context and if we trust the standard library we're guarenteeing 
// errChan will get at least 1 value.
err := <- errChan
if err != nil {
  if errors.Is(err, context.DeadlineExceeded) {
    server.Close()
    return errors.New("timed out waiting for graceful shutdown")
  }

  return err
}

return nil

File: Makefile [hunk 3]
```diff
@@ -4,6 +4,9 @@ GOARCH ?= $(shell go env GOARCH)
 # Specify the target OS to build the binary for. (Recipes: build)
 GOOS ?= $(shell go env GOOS)
 
+# Specify the GOPROXYs to use in the build of the binary. (Recipes: build, image)
+GOPROXY ?= $(shell go env GOPROXY)
```
chrisdoherty4, line 8: 🤦🏻 Thanks.

Makefile
jacobweinstock: Just trying to get the image to build at all. I found without both of these the image would fail to build.

internal/http/server.go
jacobweinstock: Apologies, I'm not following how this won't work. Both manual testing and the unit test have shown success. And it's only pseudorandom if both channels receive values at the same time. In that case it seems fine for either path to be selected.

For the shutdown, this code is exactly the same before this PR. That seems to function correctly, no?

Makefile
chrisdoherty4: We had a chat offline and clarified some of these built-in variables. It seems [BuildKit is required](https://docs.docker.com/engine/reference/builder/#automatic-platform-args-in-the-global-scope) to populate the built-in variables used in our Dockerfile. We don't document this dependency anywhere in Hegel currently.

We should, as a separate issue, figure out whether BuiltKit comes as part of the latest Docker builds and if not chose to either instrument the variables manually or document the requirement.

For now, we should be able to revert these changes and just maintain our hidden BuildKit dependency.

internal/http/server.go
chrisdoherty4: I thought on this more and I think the changes are fine as is. I was worried about missing an error if ctx.Done() is done, but frankly its irrelevant because the context is done indicating we should be shutting down so who cares if we miss the error 🤦🏻 .

chrisdoherty4 (commented):
lgtm

Just get rid of the Dockerfile changes and we're good.

jacobweinstock:
> lgtm
> 
> Just get rid of the Dockerfile changes and we're good.

I added `DOCKER_BUILDKIT=1` to the docker build command. Thoughts? I'm fine with removing this if preferred.

chrisdoherty4 (dismissed):

chrisdoherty4 (approved):

PR #194, opened 2022-11-22
Refactor for integration tests

### LINKED ISSUE #170: Kubernetes integration tests
The Kubernetes unit tests are useful for validating the conversations and how the business logic handles specific cases but it doesn't validate the logic for connecting to and retrieving hardware resources from a real API Server. 

The configuration and setup is somewhat complex, as is health checking and tear down so we should introduce some integration tests with the focus of ensuring we can talk to a real cluster.

### LINKED ISSUE #183: `/internal/http` tests have a race
The `/internal/http` tests have a race condition as it waits for the HTTP server to be deployed. Given we know we'll need some integration tests for the Kubernetes backend, it makes sense to pull this out of unit testing (given its launching a real server) and create an integration test suite that can be run independently of unit tests.

chrisdoherty4:
There are a subset of tests that require real HTTP servers, sleeps, or real external entities. These tests are being designated integration tests and will run as part of their own test suite via `make test-integration`.

This PR changes the `/internal/http` tests into integration tests and introduces Kubernetes integration tests. Additionally, it simplifies the Kubernetes backend construction by having `NewBackend` accept the `Config` object directly instead of requiring an intermediate step. 

Some additional renaming for consistency has been applied through configuration objects.

Closes #170 
Closes #183

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -32,6 +32,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
+
       - uses: actions/setup-go@v3
         with:
           go-version: "${{ env.GO_VERSION }}"
@@ -39,6 +40,23 @@ jobs:
 
       - name: Run tests
         run: go test -coverprofile=coverage.txt ./...
+
+      - name: Upload coverage report (codcov.io)
+        run: bash <(curl -s https://codecov.io/bash)
+
+  integration:
+    name: Test - Integration
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+
+      - uses: actions/setup-go@v3
+        with:
+          go-version: "${{ env.GO_VERSION }}"
+          cache: true
+
+      - name: Run end-to-end tests
+        run: make test-integration
 
       - name: Upload coverage report (codcov.io)
         run: bash <(curl -s https://codecov.io/bash)
@@ -69,7 +87,7 @@ jobs:
   e2e:
     name: Test - E2E
     runs-on: ubuntu-latest
-    needs: [test]
+    needs: [test, integration]
     steps:
       - uses: actions/checkout@v3
 
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -55,7 +55,7 @@ jobs:
           go-version: "${{ env.GO_VERSION }}"
           cache: true
 
-      - name: Run end-to-end tests
+      - name: Run integration tests
         run: make test-integration
 
       - name: Upload coverage report (codcov.io)
--- a/Makefile
+++ b/Makefile
@@ -1,3 +1,6 @@
+# Configure the Make shell for recipe invocations.
+SHELL := bash
+
 # Specify the target architecture to build the binary for. (Recipes: build, image)
 GOARCH ?= $(shell go env GOARCH)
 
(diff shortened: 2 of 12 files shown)


internal/backend/kubernetes/config.go
```diff
@@ -4,52 +4,27 @@ import (
...
-type Config struct {
-	// Config is the Kubernetes client configuration.
-	*rest.Config
-
-	// Namespace restricts the scope of the backend such that Hardware objects are retrieved from
-	// this namespace only. Defaults to "default".
-	Namespace string
-
+// BackendConfig used by the NewBackend function family.
+type BackendConfig struct {
```
chrisdoherty4, line 10: This configuration object has been simplified such that it can be passed straight to `NewBackend`. `NewBackend` now constructs the rest client provided `Config.ClientConfig` is unspecified.

Makefile
```diff
@@ -31,6 +34,30 @@ test: ## Run unit tests.
...
+# E.g. make ENVTEST_KUBE_VERSION=1.24 test-integration
+ENVTEST_KUBE_VERSION 	?= 1.25
+
+.PHONY: setup-envtest
+setup-envtest:
+	@echo Installing Kubernetes $(ENVTEST_KUBE_VERSION) binaries into $(ENVTEST_BIN_DIR); \
+	$(SETUP_ENVTEST) use --bin-dir $(ENVTEST_BIN_DIR) $(ENVTEST_KUBE_VERSION)
+
+# Integration tests are located next to unit test. This recipe will search the code base for
+# files including the "//go:build integration" build tag and build them into the test binary.
```
jacobweinstock, line 51: What are your thoughts on Peter Bourgon? https://peter.bourgon.org/blog/2021/04/02/dont-use-build-tags-for-integration-tests.html
Sorry, I don't mean to throw a wrench into this work you've done.

chrisdoherty4: I've not seen that article post before, but it was interesting. 

I'm not opposed to not using build tags; I've felt the problems the article talks about. I'm also not sure how bothered I am because the Makefile target documentation needs to clarify expectations irrespective of the approach used. The other benefit is the explicit log message which could be a signal for the ignorant to investigate further - we lack that entirely with the tags approach.

That said, I'm not a fan of skipping the test based on missing config. I'd rather have a single toggle "RunSuiteA" and the test fails for any missing required config. I feel like that's much more intentional.

jacobweinstock: what does a toggle look? Also, im good with this as is. We have some discussion items but nothing so pressing as to block this.

jacobweinstock (approved):

Makefile
chrisdoherty4: The toggle would be the same as in that article. It would replace the `is config present`. For example:

```go
func TestIntegration(t *testing.T) {
	if _, ok := os.LookupEnv(EnableIntegrationTestsEnvName); !ok {
		// This message can be more specific to the integration test suite.
		t.Skipf("To enable the integration test suite set %s=true", EnableIntegrationTestEnvName)
	}

	// Check if integration test configuration is set.
	if integrationTestConfigurationIsNotSet() {
		t.Fatal("Missing required config")
	}
}
```

I prefer this because it accommodated multiple configuration variables and feels more intentional in terms of 'turning on the test suite'.

.github/workflows/ci.yaml
```diff
@@ -43,6 +44,23 @@ jobs:
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+
+      - uses: actions/setup-go@v3
+        with:
+          go-version: "${{ env.GO_VERSION }}"
+          cache: true
+
+      - name: Run end-to-end tests
```
chrisdoherty4, line 58:
```suggestion
      - name: Run integration tests
```

Makefile
chrisdoherty4: There's quite a few PRs open with various changes so I'm going to stick with this approach for the PR but we can continue the discussion.

jacobweinstock (approved):

PR #199, opened 2022-11-26
Dynamically construct EC2 static routes

chrisdoherty4:
The EC2 frontend is composed of data routes and static routes. Static routes are those that serve child elements intead of dynamic data. They were previously manually defined which isn't particularly extensible. This change dynamically determines what static routes should exist based ont he dynamic paths.

Existing EC2 unit tests continue to pass. There was a bug in the EC2 unit tests that failed to append a trailing slash for a top level path component. E.g. `/2009-04-04` should output `meta-data/` but instead was outputting and testing for `meta-data`. It has been fixed.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -26,6 +26,8 @@ require (
 	sigs.k8s.io/controller-runtime v0.13.1
 )
 
+require github.com/kr/pretty v0.3.1 // indirect
+
 require (
 	github.com/PuerkitoBio/purell v1.1.1 // indirect
 	github.com/PuerkitoBio/urlesc v0.0.0-20170810143723-de5bf2ad4578 // indirect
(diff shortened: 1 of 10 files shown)


internal/frontend/ec2/internal/staticroute/builder.go
```diff
@@ -0,0 +1,92 @@
...
+		endpoint = "/" + endpoint
+	}
+
+	// Split the endpoint into its components so we can build the pieces we need.
+	split := strings.Split(endpoint, "/")
+
+	// Iterate over the components in reverse order so we can build parent paths for every
+	// level of path nesting and track the child part.
+	//
+	// We also need to accommodate the final "/" case. To accommodate the "/" case we do some
```
chrisdoherty4, line 37: Remove this

jacobweinstock (approved):

PR #241, opened 2023-03-15
Install tools instead of relying on `go run`

chrisdoherty4:
`go run`ning tools is slow as Go recompiles the program with ever invocation. This PR changes to installing tools into `$PWD/out/tools` ready to be used by the rest of the `Makefile` improving the dependent recipe's performance.


chrisdoherty4:
WIP: Need to work out the best way to exclude the tools.go.

Makefile
```diff
@@ -13,10 +13,70 @@ GOPROXY ?= $(shell go env GOPROXY)
...
+$(MOCKGEN): PLATFORM_ARCH := $(if $(filter x86_64,$(PLATFORM_ARCH)),amd64,$(PLATFORM_ARCH))
+$(MOCKGEN): PLATFORM_OS := $(shell echo $(PLATFORM_OS) | tr '[:upper:]' '[:lower:]')
+$(MOCKGEN):
+	curl -sSfL https://github.com/golang/mock/releases/download/$(MOCKGEN_VERSION)/mock_$(MOCKGEN_VERSION:v%=%)_$(PLATFORM_OS)_$(PLATFORM_ARCH).tar.gz | \
+		tar -C $(TOOLS_DIR) --strip-components=1 -xzf -
+	@rm $(TOOLS_DIR)/README.md $(TOOLS_DIR)/LICENSE
+
+SETUP_ENVTEST := $(TOOLS_DIR)/setup-envtest
+$(SETUP_ENVTEST):
+	go build -tags=tools -o $@ sigs.k8s.io/controller-runtime/tools/setup-envtest
```
chrisdoherty4, line 53: I realized we can just do `GOBIN=<dir> go install ...` which would align better with all other tool installations and get rid of the `tools.go`.

jacobweinstock (approved):

PR #257, opened 2023-05-10
Auto approve dependabot PRs

chrisdoherty4:
## Summary

Dependabot raises lots of updates (which is good). We require PRs to be approved before they can be merged. This auto approves dependabot PRs and leaves it to maintainers to handle the merge.


jacobweinstock:
Looks like this file needs linting, @chrisdoherty4

PR #285, opened 2023-08-22
Clarify buildkit requirement in docs

### LINKED ISSUE #196: Explicitly declare or revert BuildKit dependency
In #193 we found that BuildKit is required (https://github.com/tinkerbell/hegel/pull/193#discussion_r1029890015). We don't explicitly document this anywhere which is confusing for maintainers. 

TL;DR our Dockerfile requires some built-in build args such as TARGETOS that are populated only when using BuildKit. See https://docs.docker.com/engine/reference/builder/#automatic-platform-args-in-the-global-scope for more details.

We need to either document the requirement or ensure these variables are manually populated.

chrisdoherty4:
Closes #196

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,12 +37,17 @@ do not hesitate to raise an issue.
 
 ## Quick Start
 
+**Pre-requisits**
+- Make
+- Go
+- Docker with BuildKit
+
 ```sh
 # Build a Docker image for the host platform.
 $ make image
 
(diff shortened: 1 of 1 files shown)


chrisdoherty4:
@merigfyio queue

jacobweinstock (approved):

chrisdoherty4:
@mergifyio queue

